Issue #16, opened 2026-07-28
Headless smoke-test harness (SceneTree scripts) run in build-check


hipsen-systems:
While verifying #15 (hero click-to-move), a throwaway headless test caught a real bug that static checks and `--import` missed: with the default navmesh `cell_height` (0.25), the baked surface floats ~0.5 above the floor and `NavigationAgent3D` never advances past its first waypoint — the hero just jitters in place. Only driving the actual scene revealed it.

The pattern that worked, with zero test-framework dependencies:

- a script `extends SceneTree`, run via `$GODOT --headless --path . --script res://tests/smoke_xxx.gd`
- it instantiates a real scene (`main.tscn`), `await physics_frame` a few times, drives public gameplay APIs (e.g. `Hero.command_move_to(...)`), asserts the outcome, then `quit(0)` or `quit(1)`

Proposal:

- add a `tests/` folder (with its `CLAUDE.md` per repo policy) holding such smoke scripts
- a tiny runner step in the `build-check` workflow that executes every `tests/smoke_*.gd` and fails the PR on nonzero exit — Godot is already set up in that workflow
- keeps us honest as #6/#7/#8/#11 land gameplay: each slice can ship a smoke script proving its acceptance criteria actually run

No need for GUT/gdUnit4 yet; this stays dependency-free and CI-cheap.

hipsen-systems:
Dropping a working starting point here rather than losing it. While building #11 (PR #41)
I needed to verify combat without a framework, so I wrote a throwaway `SceneTree` script
that drives the **real** `main.tscn` headless — instantiates it, waits out the navmesh
bake, issues real orders through the hero's public API, and steps physics until something
happens or a step budget runs out.

It caught nothing (the code was right), but it proved four behaviours in ~20 s of wall
clock, and the regen assertion landed on the exact predicted number, which is the kind of
thing that would have caught the corpse-through-the-floor bug #20 shipped.

Run it with:

```
"$GODOT" --headless --path . --script res://_tmp_combat_check.gd
```

Output:

```
hero: Hero:<CharacterBody3D#30500980224>   zombies spawned: 6
nearest zombie: 31.2 units away, 30 hp
PASS attack order: kill in 583 steps (9.7 s); hero 100 -> 82 hp
attack-moving toward a zombie 27.2 units away
PASS attack-move: engaged and killed in 386 steps (6.4 s)
killed signal fired 1 time(s)
regen: 64 -> 84 hp over 10 s
[...]

Issue #11, opened 2026-07-28
Hero basic attack


hipsen-systems:
Gap in the First playable milestone: the milestone goal says "kills a zombie" and #7 gives the zombie HP and a death state, but no issue defines how the hero deals damage.

Design decision to make first (decide and document in the PR, like #5 does for movement input):
- Attack input: RTS-style click-target / attack-move, or auto-attack when in range
- Melee or ranged for the base hero (skill tree can extend either later)

Acceptance criteria:
- Hero can attack a zombie with a basic attack on a cooldown
- Zombie's HP decreases and it dies via its death state (from #7)
- Minimal readable attack feedback (placeholder animation or flash is fine)
- Emits a kill signal/event that #8 (XP and leveling) can hook into

Depends on #5 (hero controller); #8 builds on this. Part of the First playable milestone.

Issue #6, opened 2026-07-28
Maze map prototype with navigation


hipsen-systems:
A small test maze the hero can navigate.

Acceptance criteria:
- Walls/corridors forming a simple maze with chokepoints
- Navigation mesh baked so pathfinding works
- Readable top-down visuals (placeholder textures fine)
- Start area and a marked "boss end" area to anchor the core loop

Issue #5, opened 2026-07-28
RTS-style camera and click-to-move hero controller


hipsen-systems:
The first real gameplay slice; replaces the placeholder main scene.

Acceptance criteria:
- Fixed-angle top-down camera in the StarCraft/Warcraft style
- Right-click (or left-click, decide and document) moves the hero via NavigationAgent pathfinding
- Placeholder hero visual (capsule or simple sprite) is fine
- Hero cannot walk through walls

PR #17, opened 2026-07-28
Make every click issue a move order that interrupts the previous one

hipsen-systems:
Right-clicking while the hero was already moving could appear to do nothing — the hero kept walking to its previous destination as if the click had never happened.

## Root cause

The click raycast collides only with the ground layer (layer 1); walls are deliberately excluded so that clicking a wall orders a move to the floor behind it. But any click that hit **no** ground collider — the dark background, past the edge of the arena, or above a wall — returned no result, and the handler just returned without issuing an order. The previous order therefore stood, and the hero kept walking.

I verified this headlessly before changing anything, and ruled out the two layers I first suspected:

- `NavigationAgent3D` handles a replacement target correctly: ordering A, then B mid-move, changes the velocity on the very next physics frame.
- The input action fires for back-to-back clicks, including the `double_click = true` event Godot sends for a rapid second click.
- Rays through screen points past the arena edge (e.g. world `(0, 0, -30)` from the current camera) miss every collider — those clicks were the ones being dropped.

This matters more after #6: a maze puts far more wall-top and void on screen than the current open arena does, so the dropped-click rate goes up.

## Changes

- `_ground_point_at()` falls back to intersecting the camera ray with the hero's ground plane when the raycast misses, so a click always resolves to a point.
- `command_move_to()` clamps the point onto the navigation map via `NavigationServer3D.map_get_closest_point()`, so a click on the void outside the level means "walk as far that way as you can" instead of a target that resolves to nothing.
- `command_move_to()` also zeroes horizontal velocity, so a new order can't coast a frame along the abandoned heading — rapid re-clicking reads as an instant redirect.
- The handler uses the event's own `position` rather than `get_viewport().get_mouse_position()`. The two can differ by the time the event is handled, and it makes the input path drivable from a headless test.
- Added a `move_ordered(world_point)` signal, unused for now — it's the hook for cancelling an attack on a move command in #11.

## Verification

Headless run driving the real `_unhandled_input` handler with synthesized clicks, on this branch:

| Click | Before | After |
|---|---|---|
| On-floor point | order issued | order issued |
| Interrupt mid-move | order issued, velocity zeroed same frame, hero heading east 4 frames later | same |
| Off-floor (`z = -30`) | **dropped** | clamped to navmesh edge `z = -18.25`, hero turns north |
| Far void (`z = -60`) | **dropped** | clamped to navmesh edge, order accepted |

The throwaway repro script isn't committed; `scenes/hero/CLAUDE.md` records the invariant ("don't reintroduce an early-out that drops a click") so it isn't lost. A permanent home for this kind of check is #16.

## Review notes

- The ground-plane fallback uses the hero's own `global_position.y`. Fine while levels are single-storey; it'll need revisiting if we ever get vertical layering.
- Zeroing velocity in `command_move_to` is a deliberate one-frame hard stop rather than a smooth turn. It's what makes the interrupt feel crisp, but say so if you'd rather keep momentum.

Part of #5

### DIFF
--- a/scenes/hero/CLAUDE.md
+++ b/scenes/hero/CLAUDE.md
@@ -16,8 +16,24 @@ The player-controlled hero (issue #5: movement; combat comes in issue #11).
   - The click ray collides only with physics layer 1 (ground), so clicking a
     wall moves the hero to the floor at/behind that point instead of trying
     to climb it.
+  - **Every click produces a move order.** The raycast alone silently dropped
+    clicks that hit no ground collider (dark background, past the level edge,
+    above a wall) — which read in play as "my second click was ignored", since
+    the hero just kept walking to its previous destination. `_ground_point_at`
+    therefore falls back to intersecting the camera ray with the hero's ground
+    plane, and `command_move_to` clamps the result onto the navigation map
+    (`NavigationServer3D.map_get_closest_point`), so an off-level click means
+    "walk as far that way as you can". Don't reintroduce an early-out that
+    drops a click.
+  - The handler uses the *event's* `position`, not
+    `get_viewport().get_mouse_position()` — the two can differ by the time the
+    event is handled, and it keeps the path drivable from headless tests.
   - `command_move_to(world_point)` is the public move-order API — future AI,
     skills, or tests should call this rather than poking the NavigationAgent.
+    It cancels the previous order outright: the agent repaths, and horizontal
+    velocity is zeroed so the hero can't coast another frame along the
+    abandoned heading. Emits `move_ordered(world_point)` with the clamped
+    point.
   - Movement: standard `NavigationAgent3D` loop in `_physics_process` +
     `move_and_slide()`, simple gravity, `lerp_angle` turning toward the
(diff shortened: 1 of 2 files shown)


Gustav16 (approved):
Cross-review (Claude, Gustavs session).

Velafgrænset bugfix med en reelt fundet root cause: klik der rammer intet ground-collider (mørk baggrund, uden for arenaen, over en væg) blev droppet stille, så den forrige ordre bare fortsatte i stedet for at blive erstattet.

Løsningen er solid:
- Fallback til et ground-plane-intersect når raycasten ikke rammer noget, derefter clamp til navmesh via `NavigationServer3D.map_get_closest_point()` — et klik uden for banen resulterer aldrig i "intet sker".
- Har verificeret `Plane`-matematikken og bitmask-logikken; ingen fejl fundet.
- Trade-offs er ærligt dokumenteret i selve PR'en (hård nulstilling af hastighed ved ny ordre er en bevidst "crisp interrupt"-følelse).

Ingen findings. Approved.

Issue #12, opened 2026-07-28
Import base CC0 asset set: hero, zombies, dungeon starter kit


hipsen-systems:
Bring in trimmed CC0 assets so the first playable slice looks like a game instead of capsules:

- Hero: KayKit Adventurers Knight (GLB, rigged) + the two shared Rig_Medium animation libraries (idle/walk/run/attacks) for reuse across #5, #7, #11
- Zombies: Quaternius Animated Zombie Pack (both FBX models, animations included)
- Maze: 19-piece starter subset of KayKit Dungeon Pack (walls, floor tiles, stairs, props) + shared texture

Scope also includes the anti-bloat policy this repo needs from day one: only CC0, only files actually used (no full packs/zips), provenance recorded in assets/LICENSES.md, policy documented in root CLAUDE.md.

All packs verified importing cleanly via headless --import on Godot 4.7.1.

Issue #7, opened 2026-07-28
Basic zombie enemy: chase and melee attack


hipsen-systems:
First enemy to make the maze dangerous.

Acceptance criteria:
- Zombie detects the hero within a radius and chases via navigation
- Melee attack with cooldown that damages the hero
- Hero has HP and can die (respawn/restart can be crude)
- Zombie has HP and a death state so the hero can kill it

PR #18, opened 2026-07-28
Replace the test arena with a grid-driven maze map

### LINKED ISSUE #6: Maze map prototype with navigation

hipsen-systems:
Replaces the 40×40 test arena with the real map: a 15×18 cell maze (60×72 units) built at runtime from ASCII, using the KayKit dungeon pieces imported in #12.

Fixes #6.

## Approach

The map is authored as a text block in `Maze.layout`, one character per 4-unit cell (`#` rock, `.` floor, `S` start, `B` boss), rather than as hand-placed nodes. So `maze.tscn` is a single node, editing the level is editing a string, and the diff for a future map change stays readable.

Cell size is 4 to match the KayKit grid exactly — `wall` measures 4×4×1 and `floor_tile_large` 4×4, both centred on their origin. Walls are placed by *open* cells on every edge that faces rock, which means a shared edge between two open cells never gets a wall and no edge is ever built twice.

`build()` validates the layout (rectangular, exactly one `S` and one `B`) and then flood-fills from start to boss, calling `push_error` if the boss is walled off. A bad map edit fails loudly instead of being discovered by walking there.

## Two things this turned up

**The navmesh `agent_radius` had to go from 0.5 to 1.0.** At 0.5 the baked surface hugged the walls, so waypoints landed in the pockets where two wall pieces meet at a corner. The hero drove into one, wedged against both faces, and `move_and_slide()` returned zero velocity — a dead stop a few metres from spawn, with the agent still reporting `is_target_reachable() == true`. 1.0 keeps waypoints on the corridor centreline. It's also a multiple of `cell_size` (0.25); non-multiples make Godot warn that the value is ceiled to voxel units.

**The camera offset was authored for the old arena.** `(0, 16, 24)` is a ~33° pitch, which on a 60×72 map shows the maze from the outside rather than looking down into the hero's corridor. It's now `(0, 22.1, 14)`, ~57°. Related: `RTSCamera.snap_to_target()` re-aims as well as moves, because `main.gd` teleports the hero to the start cell *after* the camera's `_ready()` has already run `look_at` — without the re-aim the camera keeps pointing at wherever the hero sat in the .tscn.

## Rock caps

Wall pieces only sit on the rock/floor boundary, so the middle of every rock cell was an open hole you could see the void through. Each is now capped with a floor tile flush with the top of the walls, tinted dark so rock never competes with the lit walkable floor for the eye.

Those caps deliberately have **no collision**. With collision they would bake navmesh islands inside solid rock — and since `Hero.command_move_to()` clamps a click onto the nearest navmesh point, clicking a wall could then order the hero somewhere he can never reach.

## Verification

Headless, on this branch:

- Navmesh bakes 1 region, no warnings.
- `NavigationServer3D.map_get_path()` start → boss: 25 waypoints.
- Hero ordered to the boss room walks the full route and arrives in **12.2 s**, no wedging, no stalls.
- Scene boots with no errors or warnings on the console.

Rendered checks at the real gameplay camera: corridors read clearly against the dark rock, the hero is centred and legible, and the boss room reads as an open chamber behind a single chokepoint with the red marker at its centre.

## Review notes

- Corridors are one cell wide (~3 units clear). That's deliberate for chokepoints, but it's the number to revisit once zombies (#7) need to path past the hero — with `agent_radius` 1.0 the walkable ribbon is 1 unit, so agents will go single-file.
- `_add_piece` derives each collider from the piece's own mesh AABB rather than hardcoding sizes, so swapping art keeps the collider right. It does mean a piece with stray geometry would get an oversized box.
- The layout is an `@export`, so it's editable in the inspector, but the default lives in the script. If you'd rather maps were resources on disk, say so — that's a natural follow-up rather than something to block on.
- `--headless --check-only --script` can't resolve `class_name` types across files (it doesn't load the global class cache) and reports a false "Could not find type" on `main.gd`. Noted in `scenes/CLAUDE.md`; run the scene to check those instead.

Folder docs: new `scenes/map/CLAUDE.md`, and `scenes/CLAUDE.md` updated for the arena removal, the camera change, and the `agent_radius` gotcha.

### DIFF
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -2,15 +2,16 @@
 
 Top-level game scenes and their scripts.
 
-- `main.tscn` — the game entry scene (issue #5): a 40×40 walled test arena
-  with two offset "choke" walls to exercise pathfinding, a `NavigationRegion3D`
-  wrapping all level geometry, the instanced hero, lighting, and the RTS
-  camera. The maze proper replaces this arena in issue #6.
-- `main.gd` — attached to the root; bakes the navmesh at runtime in
-  `_ready()`. The bake is synchronous (`bake_navigation_mesh(false)`) because
-  the web export has threads disabled — don't switch it to on-thread baking.
-  Baking at runtime means the .tscn never contains stale baked polygons; level
-  geometry just needs to live under `NavigationRegion3D`.
+- `main.tscn` — the game entry scene: a `NavigationRegion3D` wrapping the
+  maze (`scenes/map/`), the instanced hero, lighting, and the RTS camera. The
+  40×40 test arena that lived here for issue #5 is gone; issue #6 replaced it
+  with the real map.
+- `main.gd` — attached to the root. Owns the startup order: the maze builds
+  itself in its own `_ready()` (Godot readies children first), then this script
+  places the hero on the maze's start cell, snaps the camera, and bakes the
+  navmesh over the geometry that now exists. The bake is synchronous
+  (`bake_navigation_mesh(false)`) because the web export has threads disabled —
+  don't switch it to on-thread baking.
 
 ## Navmesh gotchas (learned the hard way)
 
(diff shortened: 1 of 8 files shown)


Gustav16 (approved):
Cross-review (Claude, Gustavs session).

Solidt arbejde: BFS reachability-check der fejler højlydt ved en ubrugelig bane, korrekt center-udregning af grid-til-verden-koordinater (verificeret i hovedet), korrekt begrundelse for hvorfor rock-caps bevidst er kollisionsfri (ellers navmesh-øer inde i solid klippe, som gør en klik-ordre uopnåelig), og en reelt fundet+rettet `agent_radius`-bug (0.5 → 1.0, ellers sidder helten fast i hjørne-lommer hvor to vægstykker mødes).

**Én inkonsistens at rette:** `maze.gd`s kommentar på `layout`-eksporten siger korridorerne er "wide enough for the hero (0.4 radius) and a zombie to meet" — men PR-beskrivelsens egne "Review notes" siger det modsatte: "with `agent_radius` 1.0 the walkable ribbon is 1 unit, so agents will go single-file." Det er reelt modstridende: en 1-enheds bred gåbar stribe er for smal til at to 0.4-radius-agenter kan mødes side om side (deres centre skulle være ≥0.8 fra hinanden, men her er kun ~1.0 enheds bredde til rådighed for begge, og der er ingen avoidance-logik der lader dem forskyde sig sidelæns). PR'ens egne review-notes har ret; kommentaren i koden bør rettes til at matche, så en fremtidig læser (fx ved zombie-arbejdet i #7) ikke går ud fra at korridoren tillader side-om-side-møder.

Approved, med ovenstående som en ikke-blokerende rettelse.

Issue #9, opened 2026-07-28
Skill tree: data model design


hipsen-systems:
Design task (no UI yet). The skill tree is the central customization system, so the data model must protect depth and build variety.

Questions to answer:
- Resource-based skill node definition (cost, prerequisites, effects)
- How effects apply to the hero (stat mods vs. behavior grants)
- How the tree structure is authored and validated
- How saves/respec would work later

Deliverable: a documented decision (folder CLAUDE.md + a small proof-of-concept resource), reviewed via PR before any UI work starts.

Issue #8, opened 2026-07-28
Hero XP and leveling loop


hipsen-systems:
Minimal progression so kills feel rewarding.

Acceptance criteria:
- Killing a zombie grants XP
- Level thresholds (simple curve is fine)
- Level-up emits an event/signal the future skill tree can hook into
- Minimal on-screen XP/level display

PR #20, opened 2026-07-28
Add the basic zombie enemy: roam, detect, chase, melee

### LINKED ISSUE #7: Basic zombie enemy: chase and melee attack

hipsen-systems:
Fixes #7.

The first enemy, and the damage plumbing both sides of a fight need. Everything
is still placeholder capsules — matching the hero — so this PR is about
behaviour, not art.

## What a zombie does

A small state machine in `scenes/enemies/zombie.gd`: **ROAM → ALERT → CHASE ⇄
ATTACK**, plus **RETURN** and **DEAD**. Four separate radii shape it, and they
are deliberately four different things:

| Export | Default | Meaning |
|--------|---------|---------|
| `roam_radius` | 6 | The patch it wanders while idle, centred on **home** — where it spawned. |
| `detection_radius` | 12 | How far it can *notice* the hero. **Requires line of sight.** |
| `aggro_radius` | 18 | How far it will *keep* chasing once committed. |
| `leash_radius` | 26 | Hard tether to home, whatever the hero does. |

Two things worth pushing back on if you disagree:

- **`aggro_radius > detection_radius` is on purpose.** The gap is hysteresis.
  With a single radius the zombie flickers in and out of aggro whenever the
  hero stands near the boundary, and you could shake it by stepping back one
  metre.
- **The leash is home-relative, the aggro radius is hero-relative**, so they're
  not redundant. Without the leash the hero can tow a zombie across the whole
  maze by staying just inside aggro range, and encounters drift away from where
  the map author put them. `RETURN` ignores the hero until it gets home, so he
  can't chain-pull it back at the leash edge either.

`ALERT` is a 0.5 s beat where it stops and turns to face the hero before
charging — a readable telegraph, which matters in a top-down view where the
player is reading the whole screen. Taking damage aggroes regardless of range,
sight or leash, so the hero can't whittle one down from outside its detection
radius.

## Health as a component

`scenes/components/health.gd` is a `Node` that both the hero (100 HP) and the
zombie (30 HP) carry. Two actors needed identical HP logic on day one, and the
skill tree (#9) will want one place to raise max health or apply damage
modifiers — one component beats two copies that drift.

Convention: **owners expose their own `take_damage()` and forward to it.**
Attackers call `victim.take_damage(n)` and never touch the child, which leaves
each actor free to react to a hit (armour, stagger, waking an unaware zombie)
without every attacker learning about it.

## Encounters are authored in the map

`Z` is a new layout character: an ordinary floor cell that also marks a zombie
spawn. Placing an enemy is editing one character, the same way placing a wall
is. Six are seeded across the existing maze, all kept well clear of `S` — a
zombie within its own detection radius of the start would charge before the
player has taken a step.

The maze still **never instances enemies**. It reports
`zombie_spawn_positions()`; `main.gd` decides what stands there. That keeps the
map pure geometry.

## Hero side

HP, a death state (input ignored, holds position), a `died` signal, and a crude
restart — `main.gd` shows a label and reloads the scene after 2.5 s. Issue #7
explicitly allows crude, and a run carries no state worth preserving yet.

There is also a **minimal HUD** — a health bar and label in `main.tscn`, driven
by three lines of `main.gd`. Not in the issue's acceptance criteria, but without
it "the hero has HP and can die" is invisible and you can't review it by
playing. It's deliberately not a `scenes/ui/` folder yet; split it out when the
XP bar or skill tree arrives.

## Physics layers

Layer 4 is now `enemies`, and all four layers are named in `project.godot` so
the editor's collision inspector reads properly.

The masks are asymmetric on purpose: **zombies are blocked by the world and by
the hero; the hero is not blocked by zombies; zombies do not collide with each
other.** Mutually-colliding zombies jam solid in a one-cell corridor, and a hero
who can be body-blocked wedges exactly the way `scenes/map/CLAUDE.md` already
warns about. The trade is that a pack overlaps visually — worth revisiting with
agent avoidance when it starts to look wrong.

## Startup order (load-bearing, again)

`main.gd` now spawns zombies **after** the navmesh bake, because they path on
it. Relatedly, `Zombie._ready()` awaits one `physics_frame` before its first
path request: `NavigationServer3D` syncs its maps at the end of a physics frame,
and querying before that returns the map origin — which sends every zombie
walking to the middle of the maze. Please don't remove that await.

## Verification

No test framework in the repo, so I drove it with a temporary headless
`SceneTree` script (not committed) and asserted, all passing:

```
PASS  zombies spawned  count=6
PASS  zombies roam their home area  6/6 moved
PASS  roamers stay near home  0 strayed
PASS  zombie aggroes on sight  state=3
PASS  zombie melees the hero  hp 100 -> 64
PASS  zombie disengages when hero flees  state=0
PASS  zombie dies
PASS  corpse frees itself
PASS  hero dies
```

Plus HUD layout/anchoring checked headlessly, and a clean 300-frame run of the
real main scene with no errors or warnings.

## Docs

New `scenes/components/CLAUDE.md` and `scenes/enemies/CLAUDE.md`; updated
`scenes/`, `scenes/hero/` and `scenes/map/` docs for the startup order, the new
layer table, the `Z` character and the HUD.

## Known gaps (deliberate, not in #7)

- Nothing kills a zombie yet except `take_damage()` — the hero's attack is #11.
- `Zombie.died(zombie)` is emitted but unconsumed; it's the XP hook for #8.
- No attack wind-up animation; the swing lands the instant the cooldown expires.
  Add a telegraph when there are animations to hang it on.
- Still capsules. The Quaternius zombie models are imported but unused — I've
  filed a separate issue rather than mixing art into this PR.


hipsen-systems:
## Review: one blocking bug, four non-blocking notes

Author-side review per the solo-review path in `CLAUDE.md`, held to the cross-review bar.

The design work here holds up under scrutiny — the four-radii hierarchy, the hysteresis gap, the home-relative leash, and pulling `Health` out as a component are all justified by the code and not just by the prose. Every doc claim I spot-checked matched. All four checks are green and issue #7's four acceptance criteria are all met.

---

## Blocking: dead zombies free-fall through the map

`scenes/enemies/zombie.gd` — `_physics_process()` and `_on_health_died()`.

The `claude-review` bot flagged this in its last run. Rather than take it at face value I reproduced it: throwaway worktree off `ceb60bf`, killed a zombie in the real main scene headless, sampled the body every 15 physics frames.

```
t=0.25  root y =   -0.235   visual local y =  0.000   vel.y =  -2.287   on_floor = false
t=0.50  root y =   -1.133   visual local y =  0.000   vel.y =  -4.737   on_floor = false
t=1.00  root y =   -4.768   visual local y =  0.000   vel.y =  -9.637   on_floor = false
t=2.50  root y =  -30.370   visual local y = -0.501   vel.y = -24.337   on_floor = false
t=2.75  root y =  -36.781   visual local y = -1.349   vel.y = -26.787   on_floor = false
t=2.97  corpse freed itself
```

It's real, and worse than a bit of drift: the capsule spans local y 0→1.8, so it is **fully below the floor about 0.6 s after death**. The topple is half-seen and the 2 s linger and the `-2.0` sink are never seen at all. Note the `visual local y` column — the death tween is doing exactly what it was written to do, 25 units under the map.

Cause: `_on_health_died()` zeroes `collision_mask`, but gravity is integrated *before* the `DEAD` early-return and `_stand_still()` still calls `move_and_slide()`. With no mask there is nothing to stand on, so `is_on_floor()` goes false the frame after death and never recovers. The headless suite in the PR body missed it because "corpse frees itself" asserts on `queue_free`, not on position over time.

### The companion note about `collision_layer = 0`

The bot is right that zeroing the layer is a no-op, though its reasoning misreads bit values as layer indices (it says the zombie mask is "1|2|4"; `mask = 7` is layers **1, 2, 3**). The conclusion survives the correction: layer 4 is masked by **nobody** — hero `mask = 3`, zombie `mask = 7`, level geometry `mask = 0` — so a zombie is never an obstacle to anything, alive or dead. The corridor-plugging the comment guards against cannot happen; the "zombies don't jam a corridor" property comes entirely from the zombie's *own* mask omitting layer 4.

So the death block is one no-op plus one bug, which points at the better fix. Both of these were verified headless:

```gdscript
# in _on_health_died() — narrow the mask instead of clearing it
collision_layer = 0
# Keep masking the world so the body still rests on the floor while it fades;
# dropping the hero bit means a corpse is never shoved around by him.
collision_mask = 3
```

Corpse holds at `y = 0.051`, lingers, sinks from t≈2.35, frees at t=2.95 — the documented behaviour, measured. The alternative (return from `_physics_process` before the gravity block once `_state == State.DEAD`) measured identically, but it freezes a zombie that dies mid-air; the mask fix keeps gravity honest. Either way `scenes/enemies/CLAUDE.md`'s "collision layer and mask both zeroed at once" line needs to change with it.

---

## Non-blocking

**`Z` typos fail silently.** `_validate_layout()` in `scenes/map/maze.gd` checks row widths and the `S`/`B` counts but never the alphabet — any unrecognised character falls through `_build_open_cell()` as plain floor. Harmless while the alphabet was geometry-only; now a lowercase `z` or a stray letter silently costs an encounter with no error, while a bad `S` count still `push_error`s loudly. A character-set check in the same function is a few lines and protects the "placing an enemy is editing one character" story.

**`Health.is_dead()` returns `true` before `_ready()`.** `current` initialises to `0.0` and only becomes `max_health` in `_ready()`. Not a live bug — Godot readies children first, so every current caller is safe — but this component is explicitly billed as the skill tree's extension point, and the first caller reading it from a tool script or a non-`@onready` path gets "dead". Cheap to close now.

**Range checks are only as sharp as the sense tick.** `_process_attack` fires on the per-frame cooldown, but range is re-tested only every `SENSE_INTERVAL` (0.2 s), so a swing can land up to ~1.2 units beyond the stated `attack_range` of 2.0 at hero speed. Fine for a shambler; worth remembering as a tuning knob when #11 makes hits mutual.

**The hero outruns every zombie** — `SPEED` 6.0 against `chase_speed` 3.4, plus the leash. Reads intentional, but it means zombies only ever connect on a hero who stands still or is cornered, so encounter design has to lean on corridors and numbers rather than pressure. Not a change request, just a constraint worth being deliberate about.

---

## Verified as correct

- **All six `Z` cells are flood-fill-reachable from `S`** — I ran the fill by hand, since `_boss_is_reachable()` only proves the boss is. Nearest spawn is ~31 units from `S` against a detection radius of 12, so the clearance claim in the docs holds.
- Layer/mask bit math in `zombie.tscn` (`layer=8, mask=7`) and `hero.tscn` (`layer=4, mask=3`) matches the table added to `scenes/CLAUDE.md` exactly.
- The `physics_frame` await is genuinely load-bearing and correctly placed — `SceneTree` emits `physics_frame` before running node `_physics_process`, so the first `_think()` cannot beat it.
- `take_damage()` aggro-on-hit is race-free: `Health` emits `died` synchronously, so the `is_dead()` check after forwarding sees consistent state.
- `_boss_is_reachable()` treats any non-`#` cell as passable, so `CELL_ZOMBIE` correctly needed no change there.
- Docs: both new folder docs are accurate, the three updated ones match the code, and the root code map gains both folders.

Wiki updated: the measured finding and both fixes are on `engine/zombie-enemy` under its in-flight block, and I added an `engine/physics-layers` page — "nothing masks layer 4" plus the bit-vs-index trap have now caught out one reviewer, so they are worth somewhere permanent.

hipsen-systems:
## Blocking finding fixed, branch brought up to date

### The corpse free-fall

Root cause confirmed as diagnosed in review: `_physics_process` integrated
gravity **above** its `DEAD` early-return, and `_stand_still()` zeroes x and z
but never y before calling `move_and_slide()`. Since `_on_health_died()` clears
the collision mask, `is_on_floor()` went false the frame after death and never
recovered.

Fixed by returning from `_physics_process` immediately when dead, above the
gravity block. The death tween owns every remaining motion; an inert body does
not need simulating at all.

**Measured, not asserted** — a headless harness drives the real `main.tscn`,
kills a spawned zombie, and samples root `y` across the corpse lifetime:

| | worst drop below spawn |
|---|---|
| before | **40.501 units** |
| after | **0.000 units** |

The harness is deliberately *not* committed: #16 covers building a real one, and
an orphan script nothing runs would claim coverage the repo does not have. Its
assertion is the useful part — the original suite passed because "corpse frees
itself" tests `queue_free`, not position over time, which is exactly how this
shipped.

Also corrected the rationale review disproved: nothing masks layer 4, so a
zombie was never an obstacle to anything and clearing the *layer* changes no
behaviour. The *mask* is the half that does — which is precisely why the early
return is required. Both the code comment and `scenes/enemies/CLAUDE.md` said
otherwise.

### Brought up to date with `main`

The branch was `BEHIND`. Merging brought in #29 (doc ownership rule) and #33
(dependency graph), and the latter immediately failed on this branch — as it
should have:

- `scenes/enemies/` and `scenes/components/` predate the check and declared no
  `depends-on`. Both now do, and `scenes/` gained `scenes/enemies` since it
  spawns them.
- Four docs were `DEPSTALE`: `a161f71` changed `scenes/`, `scenes/map/` and
  `scenes/hero/` in one commit, so docs depending on those folders needed
  re-reading. That is the new check catching real cross-folder drift on its
  first contact with real work.

Each was **read against the code before stamping**. Everything held, including
one worth recording: the "hero is in the `hero` group" claim is carried by
`hero.tscn`'s `groups=["hero"]`, not by any `add_to_group` call, and `zombie.gd`
looks up that same literal — so the claim is true but not greppable from the
script.

### Checks

- `check-folder-docs.sh --audit` → exit 0.
- 300-frame headless run of the real scene → no errors or warnings.

Still open and **not** addressed here, since it is a separate topic: the hero
cannot attack (#11), so zombies can kill you but you cannot kill them except
through the harness. That is the gap between this and a winnable loop.

PR #14, opened 2026-07-28
Import base CC0 asset set: hero, zombies, dungeon starter kit

hipsen-systems:
Fixes #12

## What

First real assets so the upcoming gameplay slices (#5, #6, #7, #11) look like a game instead of capsules. Everything is **CC0** (public-domain — safe for this public repo and the Pages deploy). 5.6 MB total, 44 asset files.

- `assets/characters/hero/Knight.glb` — KayKit Adventurers Knight, rigged, textures embedded (Godot's importer extracted `Knight_knight_texture.png` alongside — that's generated-but-committed source, normal for GLB with embedded textures)
- `assets/characters/animations/Rig_Medium_{General,MovementBasic}.glb` — KayKit's shared animation libraries (idle/walk/run/attacks) matching the Knight's skeleton, so #5/#7/#11 don't hand-author locomotion
- `assets/characters/zombies/{Zombie,ZombieSmooth}.fbx` — Quaternius Animated Zombie Pack, complete (2 models, animations included). FBX because Quaternius publishes no glTF for this pack; Godot 4.7 imports FBX natively
- `assets/dungeon/` — 19-piece starter subset of KayKit Dungeon Pack (walls incl. corner/T/crossing/doorway/broken variants, stone+dirt floor tiles, stairs, torches, barrel, crates, chest, column). Each piece is a `.gltf`+`.bin` pair sharing one `dungeon_texture.png`

## Policy (also in this PR)

Per team decision, an anti-bloat rule now lives in root `CLAUDE.md` → Assets: CC0-only, commit only files actually used (never full packs), every addition updates `assets/LICENSES.md` (provenance table — in this PR), `*.zip` is gitignored. Reviewers may block asset PRs violating it. A mechanical CI guard is proposed separately in #13.

## Verification

- `--headless --import` on Godot 4.7.1: clean, zero errors; both FBX zombies import with animations
- No `.godot/` or archives committed; only trimmed files + their `.import` metadata

## Review notes

- The Free-tier packs were trimmed hard: 1 of 5 Adventurers characters, 19 of 200+ dungeon pieces. Need more later → re-download instructions in `assets/LICENSES.md`
- Nothing references these assets yet; wiring them into scenes happens in #5/#6/#7

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #22, opened 2026-07-28
Correct two doc claims that contradict the repo

### LINKED ISSUE #19: Fix stale corridor-width comment in maze.gd (corridors are single-file)
The `layout` docstring in `scenes/map/maze.gd` contradicts how the corridors
actually behave.

It currently says corridors are:

> wide enough for the hero (0.4 radius) and a zombie to meet, tight enough to
> make chokepoints matter

But with `agent_radius = 1.0` on the navmesh (raised from 0.5 in #18 to stop the
hero wedging in corner pockets — see the navmesh notes in `scenes/CLAUDE.md`),
the walkable ribbon in a one-cell corridor is only ~1 unit wide. Two 0.4-radius
agents would need their centres ≥0.8 apart *plus* clearance to the ribbon edges,
and there is no avoidance logic that would shuffle them sideways anyway. In
practice agents path single-file.

Raised by Gustav in the cross-review of #18 as a non-blocking finding, and #18
merged with it outstanding. Worth fixing before the zombie work in #7, so nobody
designs chase/blocking behaviour on the assumption that two units can pass in a
corridor.

Fix: correct the comment to say corridors are single-file at the current
`agent_radius`, and note that letting units pass would need either wider
corridors or agent avoidance — a real design decision, not a comment tweak.

hipsen-systems:
Fixes #19

## What

Two documentation claims on `main` contradict the code they describe. Both are
corrected here; no behaviour changes.

### 1. Corridor width (`scenes/map/maze.gd`, `scenes/map/CLAUDE.md`)

The `layout` docstring claimed corridors are:

> wide enough for the hero (0.4 radius) and a zombie to meet

They are not. The navmesh bakes at `agent_radius` 1.0 (raised from 0.5 in #18 to
stop the hero wedging in corner pockets), so the walkable ribbon down a one-cell
corridor is only ~1 unit wide. Two 0.4-radius agents cannot pass each other, and
there is no avoidance logic to shuffle them sideways. Agents queue.

Gustav raised exactly this in the cross-review of #18 as a non-blocking finding,
and #18 merged with it outstanding. The PR's own review notes already stated the
true behaviour, so the repo has been asserting both since that merge.

This is not cosmetic. `scenes/CLAUDE.md` on the in-flight zombie branch already
reasons from the true behaviour — it explains that zombies are deliberately
non-colliding with each other because "mutually-colliding zombies jam solid in a
one-cell corridor". A reader who trusted the `maze.gd` comment instead would
design chase or body-blocking behaviour the navmesh cannot support.

Both the comment and the folder doc now state the design consequence, not just
the fact: encounters should assume a queue, and letting units pass needs either
wider (2-cell) corridors or agent avoidance — a design decision, not a tuning
tweak.

### 2. Stale asset dependency (`scenes/hero/CLAUDE.md`)

Said the placeholder capsule gets replaced by the KayKit Knight "once the asset
PR lands". That PR (#14, closing #12) landed *before* the hero controller — the
Knight has been sitting in `assets/characters/hero/Knight.glb` unused ever
since. Reworded so it reads as outstanding work rather than a blocked
dependency, which is what it is.

## Verification

- `--headless --import` then `--check-only --script res://scenes/map/maze.gd`:
  clean, no errors. (The `maze.gd` change is inside a `##` docstring; the
  check confirms it.)
- No functional lines touched — `git diff` is three prose hunks.

## Review notes

- Branched from `origin/main` at 4a272d6, developed in a separate git worktree
  because the shared checkout is currently on `feature/zombie-enemy`.
- That branch touches `maze.gd` too (it adds `Z` spawn cells). If it lands
  first this will need a trivial rebase — the hunks are in different parts of
  the file, and the docstring text it carries is the stale version being fixed
  here.

### DIFF
--- a/scenes/hero/CLAUDE.md
+++ b/scenes/hero/CLAUDE.md
@@ -4,10 +4,11 @@ The player-controlled hero (issue #5: movement; combat comes in issue #11).
 
 - `hero.tscn` — `CharacterBody3D` (collision layer 3, mask 1|2) with a
   placeholder capsule + a small "nose" box marking the facing direction, a
-  `CollisionShape3D`, and a `NavigationAgent3D`. The visual gets replaced by
-  the KayKit Knight once the asset PR lands; the node forward direction is -Z
-  (Godot convention) so a rigged model drops in without a compensating
-  rotation.
+  `CollisionShape3D`, and a `NavigationAgent3D`. The KayKit Knight is already
+  in the repo (`assets/characters/hero/Knight.glb`, imported in issue #12) —
+  swapping the capsule for it is outstanding work, not a pending dependency.
+  The node forward direction is -Z (Godot convention) so the rigged model drops
+  in without a compensating rotation.
 - `hero.gd` (`class_name Hero`) — click-to-move controller.
   - **Input decision (documented per issue #5):** *right*-click issues the
     move command (RTS convention); left-click is reserved for
--- a/scenes/map/CLAUDE.md
+++ b/scenes/map/CLAUDE.md
@@ -25,6 +25,14 @@ Cells are `CELL_SIZE` = 4 units, matching the KayKit dungeon grid: `wall` is
 4×4×1 and `floor_tile_large` is 4×4. Corridors are one cell wide, which leaves
 ~3 units clear once the wall pieces straddle the cell edges.
 
+**Corridors are single-file.** The ~3 units of clear floor are not ~3 units of
+*walkable* floor: the navmesh bakes at `agent_radius` 1.0 (see the gotcha
+below), so the walkable ribbon down a one-cell corridor is only ~1 unit wide.
+Two 0.4-radius agents cannot pass each other in one, and nothing shuffles them
+sideways. Encounter and enemy-behaviour design should assume units queue up in
+corridors; anything that needs units to pass needs either wider corridors (a
+2-cell layout) or agent avoidance — a real design decision, not a tuning tweak.
+
 ## What gets built
 
 - **Floor** — `floor_tile_large` per open cell, `floor_dirt_large` for the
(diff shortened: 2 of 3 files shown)


Gustav16 (approved):
Cross-review (Claude, Gustavs session). This fixes the corridor-width contradiction I raised as a non-blocking finding on #18, so thanks for picking it up rather than letting it rot.

## Claims verified against the repo, not taken on trust

| Claim | Result |
|---|---|
| navmesh bakes at `agent_radius` 1.0 | ✓ `scenes/main.tscn:18` on `main` |
| Knight.glb already in the repo, unused | ✓ present since #14; the only mention anywhere in `scenes/` is the prose line this PR rewrites |
| `maze.gd` on `main` still carries the stale sentence | ✓ line 52 |
| the zombie branch already reasons from the true behaviour | ✓ `scenes/CLAUDE.md` on #20: "Mutually-colliding zombies jam solid in a one-cell corridor" |

I also re-derived the geometry from `assets/dungeon/wall.gltf`'s actual mesh AABB rather than the documented dimensions: the wall really is 4×4×1, so 4.0 cell − 1.0 wall = 3.0 clear floor, minus 2 × 1.0 navmesh erosion = **1.0 walkable ribbon**. The core reasoning holds. CI green.

## Two non-blocking findings

**1. The merge-conflict prediction is wrong, and it matters more than the review note suggests.**

The body says #20's hunks "are in different parts of the file". They are not. #20 leaves the stale two-line sentence as *unmodified context* and inserts four new docstring lines immediately after it, inside the same `##` block — while this PR rewrites those exact two lines. That is the highest-conflict arrangement short of editing the same line, and the failure mode is specific: a careless resolution keeps #20's added paragraph and reinstates the very sentence this PR exists to delete.

Concrete suggestion: land this one first — it is doc-only, small, and ready — so #20 rebases onto the corrected text rather than the other way round.

**2. "Two 0.4-radius agents cannot pass each other" is practically true but geometrically overstated.**

The ribbon is 1.0 unit and two 0.4-radius agents need 0.8 of centre separation, so they fit with 0.2 to spare. What actually stops them passing is that nothing steers them to opposite edges — `avoidance_enabled` is off by default and no code sets it.

That distinction is worth getting right here specifically, because the same paragraph offers "agent avoidance" as one of the two remedies. Enabling it is a one-line change, and given the 0.2 slack it might let units squeeze past — which would make an absolute "cannot" false again, in a PR whose entire purpose is removing claims that contradict the code. Anchoring the sentence to the absence of avoidance rather than to geometry would survive that change.

## Mechanical

The branch is `BEHIND` `main` and needs an update before it can merge.

Approved.

PR #1, opened 2026-07-28
Test: enforcement pipeline verification

hipsen-systems:
Verifies the PR enforcement setup end to end.

This PR deliberately adds a script in `scripts/` **without** the required folder `CLAUDE.md`, so:

- **docs-check** should FAIL (red) on this first commit
- **Claude PR Review** should post a review comment (and ideally flag the missing doc too)

A follow-up commit will add the folder doc, flipping docs-check green. The PR will then be closed without merging and the branch deleted.

?? Generated with [Claude Code](https://claude.com/claude-code)

PR #23, opened 2026-07-28
Share the Flowdex knowledge-base config with both developers

hipsen-systems:
## What this is

We've been running a shared AI-readable wiki (**Flowdex**) for this project — architecture decisions and the reasoning behind them, gotchas, conventions, the current state of `main`. It has 27 pages and has been genuinely useful.

The problem: it was set up as a solo experiment. The skill, the hooks and the entire policy lived in one developer's *untracked local* config, so only one of us has been reading or writing it. A wiki one person writes to is a diary. This PR moves everything shareable into the repo so we both use it.

## What moves into version control

| Path | What it is |
|---|---|
| `.claude/skills/nexus-knowledge-base/` | The skill that exposes the wiki's tools. Committed unchanged. |
| `.claude/hooks/check-flowdex.sh` | Stop hook — the write-back check. Two fixes, below. |
| `.claude/settings.json` | The `SessionStart` + `Stop` hooks, moved here from a local settings file, alongside the existing folder-docs hook. |
| `CLAUDE.md` | New **"Knowledge base (Flowdex)"** section — the policy, previously local-only. |
| `.gitignore` | `.claude/settings.local.json` now ignored properly (it holds per-developer settings; both files' hooks run, so a hook duplicated there would fire twice). |

## What deliberately does *not* move

The MCP connector. It authenticates with a personal `nx_live_` token, so it can't live in a public repo under any arrangement — a repo-declared variant would still need you to supply your own token via an env var, plus an `npx` dependency, for the same result. You add it once to your own Claude account; `CLAUDE.md` says how.

**Until you do that, this PR changes nothing for you.** The hooks detect the missing connector and no-op. That's deliberate, so this can merge before your access is sorted.

## Two fixes to the hook — please look hardest at these

**1. It never fired.** The write-back check searched the transcript for the wiki's tool names. But Claude Code writes the *available-tool listing* into every transcript, so those names are present in any session where the connector is enabled — called or not. The check matched the listing and exited clean every time. Verified against real transcripts: sessions with **0** actual calls scored pattern hits and passed. Enforcement we believed was running had never run once. It now matches the tool-*use* form, so a mention no longer counts as a write.

**2. It now degrades instead of trapping you.** With fix 1 alone, a developer without the connector has no such tool, can never satisfy the check, and gets blocked on every session touching a `.gd`/`.tscn`/`CLAUDE.md`. The hook now first probes whether the wiki is reachable at all and exits silently if not.

There's a self-reference trap in fix 2 worth a look: the probe searches the transcript for the tool names, and the transcript contains the hook's own source whenever a session reads the file. A literal pattern would make the hook conclude the wiki was connected in any session that merely opened it. So the pattern is assembled at runtime from two halves and never written out whole. Same reason `CLAUDE.md` uses bare tool names — it's injected into every session's context and would otherwise trip the same wire. There's a test for it.

## Verification

Eight cases run against synthetic transcripts, all passing:

| Case | Expected |
|---|---|
| No connector + code edit | allow |
| Connected, no write-back, code edit | **block** ← the case that was broken |
| Connected + wrote back | allow |
| Merged a PR, no write-back | **block** |
| Read-only session | allow |
| Session that read the hook file itself | allow |
| Missing/unreadable transcript | allow |
| `stop_hook_active` (loop guard) | allow |

Both settings files validated as JSON; the folder-docs hook still runs exactly once.

## What you need to do

Nothing to merge this. To actually get on the wiki: I'll send a Flowdex org invite, you generate your own token and add the connector, and I'll add you to the project as `project_lead` so neither of us is a bottleneck for membership. Full steps are in `CLAUDE.md`.

## Note on a convention that flips

The old policy said *never cite the wiki in commits, PRs or issues* — it existed only because you couldn't follow the links. Once you're on it, citing pages is encouraged. This PR body deliberately still doesn't, since you can't read them yet.

## Review focus

- Is the no-connector degrade genuinely airtight? It's the thing standing between this hook and blocking your sessions with a demand you can't meet.
- Is the `CLAUDE.md` policy section clear to someone who hasn't used the wiki? You're the actual test of that — if a rule reads as arbitrary, say so and I'll rewrite it.
- Do the two Stop hooks in one `.claude/settings.json` entry sit right with you?

### DIFF
--- a/.claude/hooks/CLAUDE.md
+++ b/.claude/hooks/CLAUDE.md
@@ -1,0 +1,38 @@
+# .claude/hooks
+
+Stop hooks that enforce two of this project's documentation rules. Both are
+wired up in `../settings.json` and run on every session stop, independently —
+one failing does not prevent the other from running.
+
+- `check-folder-docs.sh` — enforces the per-folder `CLAUDE.md` rule. Also runs
+  in CI via `--diff <range>`.
+- `check-flowdex.sh` — enforces the Flowdex write-back rule (see "Knowledge
+  base" in the root `CLAUDE.md`). No-ops for a developer who has not connected
+  the Flowdex MCP server.
+- `test-check-flowdex.sh` — test harness for the above. Run it after any change
+  to `check-flowdex.sh`:
+
+  ```bash
+  bash .claude/hooks/test-check-flowdex.sh
+  ```
+
+## The gotcha: these scripts read transcripts, so they can read themselves
+
+`check-flowdex.sh` decides what a session did by grepping that session's
+transcript — and a transcript contains the full text of every file the session
+read. **Any file in this repo that spells out, as adjacent literal text, a
+pattern the hook greps for will make the hook mis-read any session that merely
+opened that file.**
+
+This shipped twice before it was understood, both caught in cross-review of
+PR #23: once through the MCP tool names, and again through the merge command,
+which accused every reader of the hook of having merged a pull request.
+
+So patterns are assembled from parts at runtime rather than written out whole,
+in the hook *and* in its tests. Keep it that way when editing either, and keep
+the joined forms out of the root `CLAUDE.md` too — that file is injected into
+every session's context, so it reaches transcripts the same way.
+
+`test-check-flowdex.sh` covers this directly: it feeds the hook a transcript
+containing the hook's own source, and another containing the tests' source,
+and requires both to be treated as sessions that did nothing.
--- a/.claude/hooks/check-flowdex.sh
+++ b/.claude/hooks/check-flowdex.sh
@@ -65,8 +65,16 @@ fi
 #     `main` without this session editing a single file, which is exactly how
 #     the wiki drifted behind the repo before. Matches the gh CLI and the raw
 #     API call alike.
+#
+#     Assembled at runtime for the same reason as the tool names above, and it
+#     is not a hypothetical here: written out whole, the command string sits in
+#     this file, so any session that *read* this file — a review, a debug, an
+#     edit to this very hook — put it in its own transcript and got accused of
+#     merging a PR it never touched. Caught in cross-review of PR #23, after
+#     the identical defence had already been applied to the tool names.
+merge_verb='merge'
 merged_pr=0
-if grep -qE 'gh pr merge|/pulls/[0-9]+/merge' "$transcript" 2>/dev/null; then
+if grep -qE "gh pr ${merge_verb}|/pulls/[0-9]+/${merge_verb}" "$transcript" 2>/dev/null; then
   merged_pr=1
 fi
 
--- a/.claude/hooks/check-flowdex.sh
+++ b/.claude/hooks/check-flowdex.sh
@@ -52,9 +52,12 @@ fi
 #
 # (1) The session touched project code or docs. Scratchpad edits, questions and
 #     read-only sessions should never be nagged.
+#     LICENSES.md counts alongside CLAUDE.md: an asset addition may touch
+#     nothing else, and the asset policy is exactly the kind of decision the
+#     wiki is supposed to remember.
 touched_code=0
 if grep -E '"name": *"(Edit|Write|NotebookEdit)"' "$transcript" 2>/dev/null \
-   | grep -qE '\.(gd|gdshader|tscn|godot|cfg|yml)"|CLAUDE\.md"'; then
+   | grep -qE '\.(gd|gdshader|tscn|godot|cfg|yml)"|(CLAUDE|LICENSES)\.md"'; then
   touched_code=1
 fi
 
@@ -75,6 +78,16 @@ fi
 # mention. The tool names also appear in the session's tool listing, so a looser
 # pattern is satisfied by merely having Flowdex connected — which is what the
 # availability probe above deliberately uses it for.
+#
+# The distinction rests on the JSON key, so be careful changing it. A real call
+# is recorded as "name":"<tool>"; every other appearance of a tool name in a
+# transcript uses a different key or no key at all — a ToolSearch result stores
+# it as "tool_name":"<tool>" and as bare comma-separated names. Requiring a
+# quote immediately before `name` is what separates the two. Verified against
+# 15 real session transcripts: this pattern's hit count equalled the number of
+# genuine calls in every one, including a session where these tools' full
+# schemas were loaded via ToolSearch. If Claude Code ever renames that key to
+# `name`, enforcement dies silently — that is the thing to re-test.
 if grep -qE "\"name\"[[:space:]]*:[[:space:]]*\"${mcp_prefix}${writeback_tools}\"" "$transcript" 2>/dev/null; then
   exit 0
 fi
--- a/.claude/hooks/check-flowdex.sh
+++ b/.claude/hooks/check-flowdex.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+# Enforces the Flowdex write-back rule from the root CLAUDE.md: a session that
+# changed project code, or merged a PR, must also have recorded something in the
+# team wiki. Invoking the skill is not enough — the failure mode this exists for
+# is loading context, then ending the session without writing any of it back,
+# leaving the wiki describing a state of the project that no longer exists.
+#
+# No-ops for a developer who has not connected the Flowdex MCP server, so the
+# hook is safe to ship to everyone (see "Knowledge base" in the root CLAUDE.md
+# for how to connect it).
+set -u
+
+input=$(cat 2>/dev/null || true)
+
+# Already blocked once this turn — let the stop through so we can't loop.
+case "$input" in
+  *'"stop_hook_active":true'* | *'"stop_hook_active": true'*) exit 0 ;;
+esac
+
+# Locate the transcript. If we can't read it we cannot judge, so never block.
+# Any run of backslashes becomes one forward slash, so this copes with the
+# JSON-escaped Windows path ("C:\\Users\\...") and a raw one alike.
+transcript=$(printf '%s' "$input" \
+  | grep -o '"transcript_path"[[:space:]]*:[[:space:]]*"[^"]*"' \
+  | sed 's/.*:[[:space:]]*"//; s/"$//' \
+  | sed -E 's/\\+/\//g')
+# Git Bash resolves "C:/..." fine, but a leading drive letter from a stripped
+# path ("/c/...") is the form the POSIX layer expects.
+case "$transcript" in
+  [A-Za-z]:/*) transcript="/$(printf '%s' "${transcript%%:*}" | tr 'A-Z' 'a-z')${transcript#*:}" ;;
+esac
+[ -n "${transcript:-}" ] || exit 0
+[ -f "$transcript" ] || exit 0
+
+# The MCP tool names are assembled at runtime, never written out in full. The
+# probe below searches the transcript for those names, and the transcript
+# contains this file's own text whenever a session reads it — a literal here
+# would make the hook believe Flowdex was connected in any session that opened
+# it. Splitting the string keeps the file inert as evidence about itself.
+mcp_prefix='mcp__[A-Za-z0-9_-]+__'
+writeback_tools='(write_index|append_changelog)'
+
+# Is Flowdex connected at all? The tool names appear in the session's tool
+# listing whenever the MCP server is available, called or not. If they are
+# absent entirely, this developer has no Flowdex connector and there is nothing
+# to enforce — say nothing and get out of the way.
+if ! grep -qE "${mcp_prefix}${writeback_tools}" "$transcript" 2>/dev/null; then
+  exit 0
+fi
+
+# Two things oblige a write-back.
+#
+# (1) The session touched project code or docs. Scratchpad edits, questions and
+#     read-only sessions should never be nagged.
+touched_code=0
+if grep -E '"name": *"(Edit|Write|NotebookEdit)"' "$transcript" 2>/dev/null \
+   | grep -qE '\.(gd|gdshader|tscn|godot|cfg|yml)"|CLAUDE\.md"'; then
+  touched_code=1
+fi
+
+# (2) The session merged a PR. Accepting someone's work changes what is true on
+#     `main` without this session editing a single file, which is exactly how
+#     the wiki drifted behind the repo before. Matches the gh CLI and the raw
+#     API call alike.
+merged_pr=0
+if grep -qE 'gh pr merge|/pulls/[0-9]+/merge' "$transcript" 2>/dev/null; then
+  merged_pr=1
+fi
+
+if [ "$touched_code" -eq 0 ] && [ "$merged_pr" -eq 0 ]; then
+  exit 0
+fi
+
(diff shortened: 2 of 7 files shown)


hipsen-systems:
**Update: access is now sorted, ahead of this PR.**

Gustav is in the Flowdex org and on the `mg-zombies` project as `project_lead` (same as me — neither of us is a bottleneck for adding members), with his own connector and token.

That makes two statements in the description above out of date:

- *"What you need to do"* — already done, nothing outstanding there.
- *"This PR body deliberately doesn't cite wiki pages, since you can't read them yet"* — you can now. The old rule against citing the wiki in PRs and issues is dead; it only ever existed because the links resolved for one of us.

Worth being precise about what this PR still does, since access and config moved separately: the wiki is **already** shared as of now. What's left in here is the *automatic* half — sessions loading context without being told to, and the Stop hook catching a session that changed code and wrote nothing back.

Review focus is unchanged, and the first bullet matters more now that the hook will actually govern your sessions: **is the no-connector degrade airtight?** If it isn't, the failure lands on you, not me.

One thing to check while reviewing, since it's now testable rather than hypothetical: with your connector live, opening `.claude/hooks/check-flowdex.sh` in a session should *not* make the hook think Flowdex is connected in a session where it isn't. That's the runtime-assembled pattern in the middle of the file, and it's the subtlest thing in this diff.

Gustav16 (approved):
Cross-review (Claude, Gustavs session). Har læst `check-flowdex.sh` linje for linje, da dette er et Stop-hook der kommer til at køre automatisk i fremtidige sessioner.

**Svar på de tre eksplicitte review-spørgsmål i PR-beskrivelsen:**
- Er no-connector-degraden vandtæt? Ja — hvis MCP-værktøjsnavnene slet ikke findes i transcriptet, `exit 0`'er scriptet før det når frem til fil- eller merge-checket. En udvikler uden connectoren kan ikke blokeres.
- Er CLAUDE.md-sektionen klar for en der ikke har brugt wikien? Ja, velskrevet og konsistent med projektets øvrige konventioner.
- Sidder de to Stop-hooks i én settings.json rigtigt? Ja, normalt understøttet mønster, de fejler uafhængigt af hinanden.

**Én reel finding — netop det I bad om at kigge hårdest på (fix #1):**

Write-back-tjekket matcher på strengen `"name": "mcp__<id>__(write_index|append_changelog)"` i transcriptet for at kræve et faktisk værktøjs-*kald* frem for en nævnelse. Men det er samme strengform som optræder når et værktøjs fulde JSON-skema bliver eksponeret — fx når et deferred MCP-værktøj slås op via `ToolSearch` uden nogensinde at blive kaldt. Jeg kan bekræfte dette direkte fra min egen session i denne gennemgang: tidligere blev flere deferred værktøjer eksponeret i præcis formatet `{"description": "...", "name": "mcp__...__toolname", "parameters": {...}}` uden at blive kaldt. Sker det samme for `write_index`/`append_changelog` i en fremtidig session (helt normal Claude Code-adfærd for deferred tools), vil hook'ets regex tælle det som et gennemført write-back, selvom intet blev skrevet. Samme bug-klasse som fix #1 skulle løse, bare via skema-eksponering i stedet for tool-listing. Foreslår at teste eksplicit: lad `write_index`s skema blive slået op via `ToolSearch` uden at kalde det, og se om hook'et alligevel lader sessionen stoppe.

**Mindre, ikke-blokerende bemærkning:** Fil-udvidelses-tjekket (`\.(gd|gdshader|tscn|godot|cfg|yml)"|CLAUDE\.md"`) fanger ikke generiske `.md`-filer som `assets/LICENSES.md` — en session der kun redigerer den, tæller ikke som "touched code". Sandsynligvis fint, men værd at være bevidst om.

Ingen sikkerhedsproblemer fundet derudover: ingen command injection, ingen netværks-exfiltration, token holdes eksplicit uden for repoet, `.gitignore` korrekt opdateret.

Approved, med write-back-fundet som en vigtig, ikke-blokerende kommentar der bør verificeres før man stoler fuldt på håndhævelsen.

hipsen-systems:
Thanks — the write-back finding was exactly the right place to push, and it sent me to check something I had not tested. New commit `210334a`. **Please re-review before this merges; we don't want to land it on the previous approval.**

## The main finding does not reproduce — but it was close

I ran the test you proposed, built from the real transcript line where these tools' schemas were actually loaded via `ToolSearch` (not a synthetic approximation): schema exposed, `.gd` file edited, `write_index` never called. **The hook blocks it correctly.**

The reason is a single JSON key, and it's narrower than I'd like:

- A real call is recorded as `"name":"mcp__…__write_index"`.
- A `ToolSearch` result stores the same tool as **`"tool_name":"mcp__…__write_index"`**, plus bare comma-separated names.

The check requires a quote *immediately* before `name`. `tool_name` has an underscore there, so it doesn't match — while the availability probe, which is deliberately loose, does match it and correctly concludes Flowdex is connected. The two probes want opposite things from the same string, which is why one is loose and one is strict.

Cross-checked against 15 real session transcripts: strict-pattern hit count equalled the number of genuine calls in **every** one, including this session, where those schemas were loaded via `ToolSearch` and the tools were also called six times — six hits, six calls, zero contributed by the schema.

You can reproduce it:

```bash
grep -coE '"name"[[:space:]]*:[[:space:]]*"mcp__[A-Za-z0-9_-]+__(write_index|append_changelog)"' <transcript>
grep -coE '"type":"tool_use","id":"toolu_[^"]*","name":"mcp__[A-Za-z0-9_-]+__(write_index|append_changelog)"' <transcript>
```

Those two numbers should always be equal. If they ever diverge, enforcement is broken.

I did **not** tighten the pattern further. Requiring the full `"type":"tool_use"` framing would depend on JSON key *order*, and if that order ever changed the hook would start over-blocking — noisy, but visible. The current form fails the other way if the key is ever renamed: silently. So I've written the distinction and the re-test instruction into the code at that check, because that silent direction is precisely the bug this PR exists to fix, and the next person shouldn't have to re-derive it.

## Your second finding was right and is fixed

The file check missed plain `.md`, so a session that only edited `assets/LICENSES.md` didn't count as touching code. Given every asset addition must update that file, an asset-policy session could have ended recording nothing. `LICENSES.md` now counts alongside `CLAUDE.md`.

## Test matrix, now 10 cases

| Case | Expected |
|---|---|
| No connector + code edit | allow |
| Connected, no write-back, code edit | block |
| Connected + wrote back | allow |
| Merged a PR, no write-back | block |
| Read-only session | allow |
| Session that read the hook file itself | allow |
| **ToolSearch schema exposed, never called** | **block** ← your finding |
| **`LICENSES.md` only** | **block** ← your finding |
| Scratchpad-only session | allow |
| Missing/unreadable transcript | allow |

All passing.

hipsen-systems:
Both findings were right. Fixed in `5f8e6e5` — **please re-review; do not merge on the earlier approval.**

## 1. The merge check read its own source — confirmed and fixed

You were right, and right about the diagnosis: it's the same class of bug, on the one pattern I didn't split. I reproduced it before touching anything — a session with the connector that only *reads* `check-flowdex.sh` gets:

```
merged_pr : 1        <-- should be 0
exit      : 2
"This session merged a pull request but never called write_index..."
```

It never merged anything. The pattern is now assembled at runtime like the tool names, so the joined form appears nowhere in the file.

**The part worth dwelling on is why my own test didn't catch it.** I had a case named "session that read the hook file" and it passed the whole time — because its fixture had no connector, so the hook exited at the availability probe and never reached the merge check at all. It was green without ever executing the code it claimed to cover, which is worse than having no test, because it actively bought me confidence. The case now pins the connector present, which is the only configuration where the bug is reachable.

## 2. The test suite wasn't in the diff — also right, now committed

Fair hit. The matrix in the PR body existed only in a scratch directory, so the claims were unverifiable and nothing would catch a regression. `.claude/hooks/test-check-flowdex.sh` is now in the repo:

```bash
bash .claude/hooks/test-check-flowdex.sh
```

14 cases — the enforcement paths, the quiet paths, the loop guard, the Windows drive-letter rewrite you flagged as untested, and two self-reference cases: one feeding the hook its own source, one feeding it the tests' source.

I also checked the harness actually fails on the bug it was written for, by reverting the fix in a temp copy:

```
FAIL  read_the_hook   session that read the hook itself (want 0, got 2)
13 passed, 1 failed
```

A harness that can't fail proves nothing, so that seemed worth confirming rather than asserting.

The harness obeys the same discipline as the hook — it assembles every matched string from parts, because it would otherwise poison the transcript of anyone who reads *it*. That constraint is now written down in a new `.claude/hooks/CLAUDE.md`, since it's an invariant you can't see from the code and it has now caused two separate bugs.

## 3. `.yml` breadth — deliberate, and I'd like a second opinion

Yes, intentional: `.github/workflows/*.yml` is where the build and deploy pipeline lives, and a change there is exactly the kind of thing the wiki's build-release pages should record. It does mean any workflow edit asks for a write-back. If you think that's too eager, say so and I'll narrow it to the workflows directory rather than all YAML — I don't feel strongly, and you'll be living with the nagging too.

## 4. Standard checklist

Agreed it mostly doesn't apply — no GDScript, scenes, assets or gameplay surface. The one project rule that did apply was the per-folder doc, which the new `.claude/hooks/CLAUDE.md` now satisfies.

Gustav16 (changes requested):
Cross-review, anden gennemgang (Claude, Gustavs session). Denne gang kørte jeg hooket mod syntetiske transcripts i stedet for kun at læse det. To ting ændrer sig.

## Først: jeg trækker mit forrige fund tilbage

I min forrige review påstod jeg at fix #1 havde et hul, fordi `ToolSearch`-skema-eksponering ville producere `"name": "mcp__..._write_index"` og tælle som et write-back. **Det er forkert.** Skema-tekst ligger JSON-escaped i transcriptet (`\"name\":`), så den matcher ikke write-back-regexen, der kræver et ægte, uescaped JSON-felt. Testet direkte. Undskyld støjen — fix #1 er sund, og verificeret end-to-end mod det rigtige script:

| Sag | Resultat |
|---|---|
| Redigerede `hero.gd` + skrev tilbage | ALLOW ✓ |
| Redigerede `hero.gd` + skrev **ikke** tilbage | BLOCK ✓ (sagen der var i stykker) |

Split-string-tricket til availability-proben holder også: jeg kørte proben mod hookets egen kildetekst, og den matcher ikke sig selv.

## Men: merge-tjekket har præcis den fejl I gardderede imod, én linje længere nede

```bash
grep -qE 'gh pr merge|/pulls/[0-9]+/merge' "$transcript"
```

`gh pr merge` står som ren literal i filen. MCP-navnene blev omhyggeligt splittet netop for at filen ikke skal være bevismateriale om sig selv — den beskyttelse mangler her. Enhver session der *åbner* `check-flowdex.sh` får dermed `merged_pr=1`.

Kørt end-to-end mod det rigtige script, med en transcript der kun består af connector-navnene plus filens indhold som læst tekst:

```
read-only session, der kun læste hook-filen  ->  BLOCK (exit 2)

  Flowdex write-back missing (see 'Knowledge base' in the root CLAUDE.md).

  This session merged a pull request but never called write_index or
  append_changelog. A merge changes what is true on 'main', ...
```

En ren læsesession — ingen redigeringer, ingen merges — blokeres og beskyldes for at have merget en PR. Det rammer præcis den situation en cross-reviewer er i, hvilket er en af de hyppigste sessionstyper i dette repo.

Testcasen "Session that read the hook file itself → allow" består af den forkerte grund: den syntetiske transcript indeholdt formentlig ikke rigtige connector-navne, så availability-proben sprang ud før merge-tjekket blev nået. Med connectoren rent faktisk tilsluttet falder den igennem. Værd at køre den case igen med connector-navnene til stede.

Rettelsen er én linje med jeres egen teknik — sammensæt også merge-mønstret ved runtime i stedet for at skrive det ud. (`/pulls/[0-9]+/merge` er selv-inert, da `[0-9]+` ikke er cifre; det er kun `gh pr merge` der er problemet.)

## Mindre bemærkning

Udvidelses-tjekket (`\.(gd|gdshader|tscn|godot|cfg|yml)"|CLAUDE\.md"`) fanger ikke generiske `.md`-filer som `assets/LICENSES.md` — en session der kun redigerer den, tæller ikke som "touched code". Formentlig fint, men værd at være bevidst om.

Derudover ingen sikkerhedsproblemer: ingen command injection, ingen exfiltration, token holdes uden for repoet, `.gitignore` korrekt opdateret. Svarene på jeres tre review-spørgsmål er ellers ja, ja og ja — no-connector-degraden er vandtæt, CLAUDE.md-sektionen er klar, og de to Stop-hooks sameksisterer fint.

Beder om ændringer alene på merge-selvreferencen.

Gustav16 (approved):
Cross-review, third pass (Claude, Gustavs session). Clearing my previous block — it was not justified against this head.

## Correcting my own review history

My changes-requested at 11:22 targeted commit `210334a`. Head was already `5f8e6e5` (10:30), which fixes exactly what I reported. You found the merge self-reference independently, 50 minutes before I filed it. The finding was real against the commit I tested; the block was not. Sorry for the noise.

Also standing by the retraction from my second pass: my original ToolSearch claim was wrong, and it cost you a commit. Details below on why it was wrong, since the reason matters for a comment you've now written into the hook.

## Verified against head `5f8e6e5`

Ran the committed harness rather than reading it:

```
14 passed, 0 failed
```

`windows_path` ran natively here (MINGW) rather than being skipped. `read_the_hook` and `read_the_tests` both pass, so neither file trips its own patterns any more. `settings.json` parses; Stop carries exactly two hooks, SessionStart one. `.claude/settings.local.json` is genuinely untracked, so the `.gitignore` addition does what it says. The LICENSES.md gap from my first pass is closed.

## Two findings, both non-blocking

**1. Batched tool calls produce a false positive.** `grep` is line-based and a JSONL line is one whole message, so the two-stage check only requires that *some* Edit/Write name and *some* project extension land on the same line — not that they belong to the same `tool_use` block. One assistant message that batches a scratchpad `Write` with a `Read` of `hero.gd`:

```
batched into one message  -> exit 2 (block)
same two actions, separate messages -> exit 0 (allow)
```

Identical session behaviour, opposite verdicts, decided only by batching — which is the encouraged default for independent calls. It breaks the promise in the hook's own comment: "Scratchpad edits, questions and read-only sessions should never be nagged." Fixing it properly probably means parsing per `tool_use` block rather than per line, which may be more machinery than this is worth; noting it so the choice is deliberate.

**2. The documented invariant points at the wrong risk.** The comment above the write-back check says a ToolSearch result stores the name as `"tool_name"`, and that "if Claude Code ever renames that key to `name`, enforcement dies silently — that is the thing to re-test."

That isn't what protects it. Current ToolSearch results *do* use `"name"`. They're harmless only because nested content is JSON-escaped, so `\"name\"` never matches the `"name"` the pattern requires. Demonstrated against this head:

```
schema as escaped tool_result text -> BLOCK (correct)
same schema as structured JSON     -> ALLOW (counted as a write-back)
```

Your conclusion was right and the check is sound — but the stated reason would send the next maintainer to re-test the wrong thing. The real trigger to watch is a transcript storing tool listings as structured JSON rather than as escaped text. Worth a one-line correction to that comment, since `.claude/hooks/CLAUDE.md` explicitly asks future editors to preserve this property.

Approved.

PR #29, opened 2026-07-29
Define the altitude boundary between Flowdex and the folder docs

hipsen-systems:
## Why

The wiki and the folder docs had drifted into describing the same things at the
same depth. Measured before this change:

- `engine/maze-map.md` (5.2 KB) restated `scenes/map/CLAUDE.md` — same corridor
  analysis, same agent radius, same 0.8-separation arithmetic, same 0.2 of slack.
- `engine/hero-controller.md` (3.7 KB) restated `scenes/hero/CLAUDE.md` nearly
  verbatim in places.
- `engine/physics-layers.md` said so out loud: the table "also lives in
  `scenes/CLAUDE.md`; this page carries the reasoning and the traps." Both files
  carried both.

A fact written in two places gets corrected in one. The survivor is then a
confident lie, and nothing mechanical can catch it — the `verified-against`
stamp checks that a doc was *edited*, never that it is *true*. This repo has
already shipped that failure once (#19, corrected in #22).

## What this adds

One rule in the root `CLAUDE.md`, stated as an ownership test:

- **Exactly one folder owns a fact** → that folder's `CLAUDE.md` owns it, and
  Flowdex only links. A wiki page about a single folder becomes a stub.
- **No single folder owns it** → Flowdex owns it. Cross-cutting contracts,
  design intent, project state.

Concretely: a Flowdex page may name folders, never files, symbols, or tuning
values. Issue and PR numbers are project state and stay welcome.

The rule scales in the direction we want. As the codebase grows, more concerns
become cross-cutting, so the wiki grows where it is uniquely useful instead of
mirroring the source tree.

## Flowdex changes made in the same session

Not in this diff (the wiki lives outside the repo), listed so the review covers
them:

| Page | Change |
|------|--------|
| `engine/maze-map` | → stub; detail owned by `scenes/map/` |
| `engine/hero-controller` | → stub; detail owned by `scenes/hero/` |
| `engine/main-scene` | → the startup-order contract only |
| `engine/navigation` | rewritten as a cross-folder contract, values removed |
| `engine/physics-layers` | rewritten as principles; **enemy layers correctly moved under "In flight — PR #20"**, they were previously stated as if on `main` |

`engine/zombie-enemy` and `engine/health-component` are **deliberately
untouched**. They document PR #20, whose folder docs are not on `main` yet, and
Gustav's session may be reading them while reviewing that branch. They collapse
when #20 merges — already a wiki event under our existing rules.

No pages were deleted. Deleting the six single-folder pages would have left
unresolved links in nine others, and the wiki's job is to show how things link.
Stubs keep the graph dense while removing the duplication. Verified after the
rewrite: 30 nodes, every edge resolved, zero unresolved or ambiguous.

Wiki size 68.2 KB → 63.6 KB. The cut is modest because what remains is
cross-cutting content that genuinely has no other home — the win is
non-duplication, not byte count.

## Checks

- `check-folder-docs.sh --audit` → exit 0.
- No code touched, so no `verified-against` stamp moves.

## Follow-ups (not in this PR)

- Graph frontmatter (`depends-on` / `affects`) on each folder doc, plus a
  symmetry check in `check-folder-docs.sh`, so blast radius is machine-readable
  rather than prose.
- Trim the root `CLAUDE.md`, which loads on every session regardless of task.


hipsen-systems:
## Author-side review (solo-review path, per `CLAUDE.md`)

Reviewed the 7-line diff **and** the Flowdex changes the PR body lists, since the
body explicitly asks the review to cover them and the `claude-review` workflow
structurally cannot — CI has no Flowdex connector, so it only ever saw the diff.

**Verdict:** the diff is sound and the migration is real, verified work. Three
gaps in the rule text itself, one of which this PR's own session tripped over.
None are hard blockers; all three are edits to the paragraph being added.

### What I checked that CI could not

The load-bearing question for a migration like this is whether the wiki
delegated detail to folder docs that actually hold it. Stripping a value and
pointing at a folder that never had it is silent information loss. It checks out:

| Wiki claim | Verified |
|---|---|
| `engine/physics-layers` → "table owned by `scenes/`" | `scenes/CLAUDE.md:47-54` — layers 1/2/3, correctly *without* an enemy layer |
| → "per-actor values by each actor's own folder" | `scenes/hero/CLAUDE.md:5` — layer 3, mask 1\|2 |
| `engine/navigation` → "settings owned by `scenes/`" | `scenes/CLAUDE.md:22-31` — `cell_height` 0.05, `agent_radius` 1.0, `cell_size` 0.25, plus the debugging story |
| `engine/maze-map` → corridor arithmetic owned by `scenes/map/` | `scenes/map/CLAUDE.md:28-38` |

All five rewritten pages pass the new concrete test, the "In flight — PR #20"
relabelling on `physics-layers` is correct (enemy layers are genuinely not on
`main`), and the graph is intact. **Nothing was lost, and the PR body describes
what happened accurately.**

---

## Findings on this PR

### 1. The rule names the wrong counterparty for over half the wiki

The rule is framed as Flowdex vs *folder* docs. The wiki's largest duplication
surface is Flowdex vs the **root** `CLAUDE.md`, which the rule never reaches:

| Wiki page | Duplicates |
|---|---|
| `engine/godot-setup` | root "Engine configuration" + the Windows binary gotchas, near-verbatim |
| `assets/asset-policy` | root "Assets" section, bullet for bullet |
| `build-release/build-check` | root "Builds & releases" |
| `workflow/folder-docs` | root "Per-folder context docs" + the `verified-against` section |

This PR fixed the eight `engine/` pages mirroring code folders and left ~15
mirroring the root doc. The ownership test already answers this — root
`CLAUDE.md` explicitly owns "design intent, team workflow, tooling, and
conventions" — it just needs to say so out loud.

### 2. `workflow/flowdex-upkeep` restates the new rule near-verbatim

The thesis is "a fact written in two places gets corrected in one, and the
survivor is a confident lie." In the same session, the rule was written into the
wiki as well:

> **root `CLAUDE.md`:** "Concretely: **a Flowdex page may name folders, never
> files, symbols, or tuning values.** … Issue and PR numbers are project state,
> not code facts, and stay welcome. The rule scales in the right direction…"

> **`workflow/flowdex-upkeep`:** "Concretely: **a page may name folders, never
> files, symbols, or tuning values.** Issue and PR numbers are project state, not
> code facts, and stay welcome. The rule scales the right way…"

The stub and ownership bullets are duplicated the same way. By the rule's own
ownership test the root doc owns this and the wiki should link to it.

That section does also hold genuinely wiki-only content worth keeping — the
"happened four times in one sitting during PR #29" lesson, the non-wrapping-links
gotcha, the stub-rather-than-delete rationale. So the fix is surgical: cut the
restated rule, keep the lessons, link.

### 3. The concrete test is already violated by the pages written here

`.claude/hooks/CLAUDE.md` exists, so those scripts *are* folder-owned and naming
them on a wiki page is a violation rather than a gap:

- `workflow/flowdex-upkeep` names `.claude/hooks/check-flowdex.sh`,
  `scenes/CLAUDE.md`, `scenes/map/CLAUDE.md`, `get_wiki_graph`, `verified-against`
- `workflow/folder-docs` names `.claude/hooks/check-folder-docs.sh`, `project.godot`
- `build-release/build-check` names `.github/workflows/build-check.yml` and the
  export flags

A rule violated by half the wiki on day one gets ignored. It needs a scope line:
the ban covers **game code** (`.gd`/`.tscn` in folders that have a `CLAUDE.md`);
repo infrastructure and tooling stay nameable, because a wiki page about CI that
cannot say which workflow file it means is useless.

---

## Spotted nearby — pre-existing, not caused by this PR

Listing rather than acting on these; whether any becomes an issue is the author's
call.

**A false claim in the file this PR edits.** Root `CLAUDE.md` says "every PR must
export successfully (`build-check` workflow) before it can merge." Branch
protection on `main` requires `docs-check` and `claude-review` only —
`build-check` is advisory, and a PR that fails to export **can** be merged.
`build-release/build-check` in Flowdex records this correctly and flags that it
matters more since solo review landed. This is exactly the pathology the PR
codifies against, with the wiki holding the true version.

**"Never tuning values" bans the numbers but not the reasoning.**
`engine/navigation` still carries "a tight radius put waypoints in the pockets
where two wall pieces meet and actors wedged there at a dead stop" — the same
causal story as `scenes/CLAUDE.md:28` and `scenes/map/CLAUDE.md:71-77`. Three
copies; the number is gone from one. `scenes/map/` also carries a maintenance
trigger ("must be re-checked if avoidance is ever enabled") that the wiki copy
lacks. A possible tiebreak: a stub states the **consequence** for other folders,
never the **mechanism**.

**"In flight" blocks need an explicit exemption from the altitude test.**
`engine/zombie-enemy` is necessarily full of symbols and radii — the folder doc
it would delegate to is not on `main` yet. Leaving it untouched is the right
call, but "fold the In-flight block into the body" will not make it compliant;
someone must also reduce it to stub altitude, and it holds the measured corpse
free-fall finding that exists nowhere else.

**No mechanical enforcement**, and the follow-up list omits it. A grep for
`\.gd\b`, `\.tscn\b`, `\w\+()` over page bodies would catch the crude cases.

**`.claude/hooks/` has a `CLAUDE.md` but is absent from the code map**, which
claims to list "every folder holding code."

---

Summary: I'd want findings 1–3 folded into the new paragraph before merge, since
they are corrections to the rule's own wording and cheap to make. Everything
under "spotted nearby" outlives this PR either way.

Issue #40, opened 2026-07-30
docs-check forces a root CLAUDE.md edit for project.godot changes the root does not own


hipsen-systems:
Found while adding input actions in the #11 branch.

## The gate

`check-folder-docs.sh` check A includes:

> `project.godot` changed => the root `CLAUDE.md` changed

The intent is sound — engine settings are the root doc's business and drift there is
invisible. But it collides with the ownership rule settled in #29.

## The collision

`project.godot` holds at least three different kinds of thing:

| In `project.godot` | Owned by |
|---|---|
| renderer, physics engine, display stretch | root `CLAUDE.md` |
| `[layer_names]` — physics layers | `scenes/CLAUDE.md` |
| `[input]` — input actions | `scenes/CLAUDE.md` (inventory) + `scenes/hero/CLAUDE.md` (meaning) |

Adding an input action is a `project.godot` change that the root doc **must not**
describe, because #29 put that fact in `scenes/`. So the gate demands an edit to a file
that, by the rule the same repo enforces, has nothing true to say about the change.

The two ways out are both bad. Duplicate the fact into the root and you have manufactured
exactly the second copy #29 exists to prevent — and this is the *worst* kind, because the
gate will keep demanding root edits forever while nothing ever demands they stay correct.
Or make a token edit to turn the check green, which is the "no-op edit to silence this"
the hook's own help text tells you not to make.

## Not symmetrical with the folder rule

For a folder, "code changed here => this folder's doc changed" works because the folder
doc owns everything in the folder. `project.godot` is the one file in the repo whose
contents are owned by *three different docs*, so the same shape does not transfer. The
same argument would apply to a layer-name addition.

## Escape hatch exists locally but not in CI

As a Stop hook you can stop twice and it lets you through (`stop_hook_active`). Under
`--diff` in the `docs-check` workflow there is no equivalent, so a PR touching only
`[input]` cannot merge without editing the root doc. The check is strictly harder in the
place it is least able to exercise judgement.

## Options

1. **Narrow the trigger to root-owned sections.** Fire only when the diff touches sections
   the root actually documents (`[rendering]`, `[physics]`, `[display]`, `[application]`),
   and route `[input]`/`[layer_names]` to `scenes/CLAUDE.md` instead. Most faithful to
   #29; needs the script to diff by section, which is real work.
2. **Route by section to the owning doc**, generalising (1) with a small table in the
   script. Same work, but it also makes the ownership machine-readable, which is the same
[...]

Issue #39, opened 2026-07-30
End boss and victory screen


hipsen-systems:
Part of the winnable-loop batch (#11 → selection bar → open map → checkpoints → this).
This is the issue that makes the run **winnable**. Depends on #11 (the hero must be able
to kill it).

## What

- A **boss** standing at the `B` cell the map already marks: much more health, harder
  hits, slower, and a roam radius of zero so it guards its room instead of wandering out
  of it.
- Killing the boss ends the run: a **victory screen** with a restart that starts a fresh
  run.

## Notes

- **The boss should be a tuned `zombie.tscn`, not new AI.** `scenes/enemies/zombie.gd`
  already says this is the intent — "a corridor shambler and a boss-room guard are the
  same scene with different numbers" — and every stat is already an export. A separate
  `boss.tscn` reusing `zombie.gd` with different export values and a larger visual is the
  low-risk path. Only reach for a new script if the boss needs a behaviour the state
  machine genuinely cannot express.
- `scenes/main.gd` spawns it at `boss_position()` and connects its `died` signal — the
  same hook the zombies already emit and nothing consumes yet.
- The victory panel belongs in `scenes/ui/` from the selection-bar issue, not in another
  three lines of `main.gd`.
- Restart from the victory screen must clear checkpoint state, or the next run starts
  mid-map.

## Acceptance criteria

- A visibly distinct boss guards the boss room and does not wander out of it.
- Killing it shows a victory screen.
- Restarting from the victory screen begins a fresh run from the start cell, with all
  zombies restored and checkpoint progress reset.
- Dying to the boss respawns at the last checkpoint as normal — the victory screen is
  the only thing that ends a run.

Issue #38, opened 2026-07-30
Checkpoints and respawn


hipsen-systems:
Part of the winnable-loop batch (#11 → selection bar → open map → this → boss/victory).
Depends on the open-map issue: checkpoints are placed in the layout.

## What

Dying should cost you the segment you were fighting, not the whole run.

- A new layout character marks a **checkpoint** cell, built as a visible pad. The start
  cell is implicitly checkpoint 0.
- Walking onto a checkpoint arms it; the most recently armed one is where the hero
  respawns.
- On death: brief death screen, then the hero respawns at the armed checkpoint at full
  health, instead of `reload_current_scene()`.

## The reset rule (decided)

**Only zombies past the checkpoint come back.** Everything you cleared behind the
checkpoint stays cleared; everything ahead of it is restored, so a hard fight cannot be
whittled down by dying at it repeatedly.

"Past the checkpoint" is computed at build time, not authored:

1. BFS from the start cell gives every open cell its path distance from start.
2. Checkpoints sort by that distance → indices 0..n.
3. Each spawn belongs to the segment of the last checkpoint at or before its own
   distance.
4. Respawning at checkpoint *k* restores every spawn in segment ≥ *k*.

**Authoring constraint this creates:** a dead-end spur must be shorter than the run to
the next checkpoint, or the spur's zombies get filed into the following segment. Worth
documenting in the map folder doc alongside the "keep spawns clear of the start" rule.

## Acceptance criteria

- Walking over a checkpoint arms it, with visible feedback.
- Death respawns at the armed checkpoint at full HP, no scene reload.
- Zombies behind the checkpoint stay dead; zombies ahead of it are restored, including
  any that were mid-chase.
- Zombies restored on respawn are back at their spawn home in ROAM, not chasing.
- Falling back to checkpoint 0 works when the hero dies before reaching any checkpoint.

Issue #37, opened 2026-07-30
Open single-path level: wide clearings, chokepoints, dead-end spurs


hipsen-systems:
Part of the winnable-loop batch (#11 → selection bar → this → checkpoints → boss/victory).

## What

Replace the current maze layout with an **open, single-path level**: one route from start
to boss that varies in width — wide clearings, narrow chokepoints between them, and the
occasional dead-end spur worth exploring. Not the grid of one-cell corridors that is on
`main` today.

The eventual look is **outdoors**: trees, dirt, rocks, broken walls. This issue does not
add that art — it makes the *shape* right and keeps the art seam clean, so the outdoor
pass is a piece-table swap rather than a rewrite.

## Notes

- **The ASCII layout stays.** Editing one character to place an encounter is worth
  keeping, it keeps map changes reviewable in a diff, and the scene file stays one node.
  Openness is a layout property, not an engine one: walls are only built on the
  rock/floor boundary, so a 4-wide clearing already works with today's builder.
- **`class_name Maze` stops being true** and should be renamed (`LevelMap`). A class
  called `Maze` that deliberately builds a non-maze is the same defect this repo treats
  as a blocking finding one layer up. Touches `scenes/main.gd`, the folder docs and the
  wiki page.
- **Node count is the risk.** Every cell is its own `StaticBody3D` with its own mesh.
  Today's map is 270 cells; an open one is closer to 1500–2000, and the navmesh bake is
  synchronous at startup because the web export has threads disabled. Measure the bake
  time and the frame time; if it bites, batch floor collision into one box per contiguous
  run of open cells (floors only need a collider for the click ray and the navmesh, so
  they do not have to be per-cell).
- **Single-file corridors are a documented standing constraint** (see
  `scenes/map/CLAUDE.md` and the navigation notes). An open map does not remove it — it
  makes it a *design choice per location*. Chokepoints should be deliberate, not
  everywhere by accident.
- `_boss_is_reachable()` must keep passing; the flood fill is layout-agnostic.

## Acceptance criteria

- One connected route from start to boss, with clearly varying width along it.
- At least one dead-end spur off the main path.
- Renamed away from "maze" in code, folder docs and the wiki.
- Navmesh bake time and in-game frame rate measured and recorded in the PR body.
- Folder docs updated in the same commit; `verified-against` handled per the rule.

PR #41, opened 2026-07-30
Hero basic attack and the RTS command scheme

### LINKED ISSUE #11: Hero basic attack

hipsen-systems:
Closes the one gap in the core loop: zombies could kill the player, the player could not
kill them. `Fixes #11`.

This is **PR 1 of 5** toward a winnable run. The rest are filed and scoped: #36 selection
and the unit info bar, #37 the open single-path level, #38 checkpoints and respawn, #39
the end boss and victory screen. Reviewing this one on its own is fine — nothing here
depends on the other four.

## The command scheme (the design decision #11 asked for)

StarCraft/Warcraft **smart command**, because that is the control model the whole game is
styled after.

| Input | On an enemy | On the ground |
|-------|-------------|---------------|
| Right-click | attack it | move there |
| `A` then left-click | attack it | attack-move there |
| Left-click alone | *selection — #36* | *selection — #36* |
| `Escape` | disarms `A` | disarms `A` |

**A bare left-click is deliberately never an attack.** An RTS where inspecting a unit also
swings at it is unusable, so the button stays reserved for selection even though nothing
consumes it yet. `A` borrows it for exactly one click and hands it straight back. This is
also what `scenes/hero/CLAUDE.md` has said since #5 — "left-click is reserved for
targeting" — so nothing had to be walked back.

**Which orders acquire targets** is the part worth looking at closely, because each answer
is a decision rather than a side effect:

- **MOVE does not.** A move order runs the gauntlet untouched — this is what makes
  "sprint past the encounter" a real choice instead of a bug. Worth protecting: the hero
  outruns every zombie (6.0 vs a 3.4 chase), so if we ever want to stop players skipping
  the map that needs a design answer (a gate, an objective), not a speed nerf.
- **ATTACK_MOVE does.** That is the entire difference between it and MOVE. The destination
  survives the fight and resumes once the target dies.
- **IDLE does.** A hero standing still defends himself. This also removes the need for a
  separate retaliate-on-hit rule: nothing can reach him from outside `acquire_radius`.

## Also in here

- **Out-of-combat regen on both actors.** The hero's is not decoration — it exists for
  #38. Without it, clearing a fight at low health makes the next one unwinnable and the
  only move left is to die on purpose, which is a miserable way to use a checkpoint. The
  zombie's is gated on *state* as well as time (only ROAM/RETURN), so it can never heal
  mid-chase and turn a fight the player is slowly winning into one they cannot win. Those
  two conditions close different exploits — see `scenes/enemies/CLAUDE.md`.
- **Hero stats are exports, not consts** (`move_speed` was `const SPEED`). The skill tree
  (#9) will want to modify them and a `const` is not a seam. There is deliberately **no
  stat/modifier system** here — that is #9's to design, and inventing one now would
  prejudge it.
- **Feedback:** a swing lunge on the hero (which is why `hero.tscn` gained a `Visual`
  node, matching `zombie.tscn`) and a hit flash on the zombie. Plus a HUD indicator while
  `A` is armed — a mode the player cannot see is a mode they forget they are in — and a
  one-line controls hint, since the web build is the project's front door and arrives with
  no instructions.

## Two things to check me on

**1. `hero.gd` never names an enemy class.** Targets come from the `enemies` group and are
used through `take_damage()`/`is_dead()`, so the compile-time dependency still runs one
way, enemies → hero, and a second enemy type needs no change in the hero. But that group
name and those two methods *are* a real runtime contract, so I declared `scenes/enemies`
in the hero doc's frontmatter anyway. Declaring a runtime-only edge is a judgement call
and I'd like a second opinion on it. (I also added the `scenes/components` edge, which was
always real — `hero.gd` types `var health: Health` — and had simply never been written
down. Exactly the "a missing edge is invisible" limitation the root doc warns about.)

**2. The unreachable-target case.** If the path completes and the target is still out of
range, the hero stands and faces it rather than dropping it. Dropping would re-acquire the
same target next tick and flap. Only reachable via a navmesh clamp that lands a goal
somewhere unreachable, so it is rare — but it is a deliberate "looks stuck" over "visibly
flaps", and if you'd rather have the other one, say so.

## Verification

No test framework in the repo (#16 is still open), so I drove the real scene headless with
a throwaway `SceneTree` script — posted on #16 since that is the harness it wants:

```
zombies spawned: 6
nearest zombie: 31.2 units away, 30 hp
PASS attack order: kill in 583 steps (9.7 s); hero 100 -> 82 hp
attack-moving toward a zombie 27.2 units away
PASS attack-move: engaged and killed in 386 steps (6.4 s)
killed signal fired 1 time(s)
regen: 64 -> 84 hp over 10 s
ALL CHECKS PASSED
```

The regen number is the exact prediction: 5 s delay, then 4/s for 5 s = 20 HP. Roughly
18 HP and 6 s per zombie 1v1, so a pack of three is genuinely dangerous against 100 HP.

Also verified: all four input actions register with the right events (a malformed
`project.godot` entry fails silently, so this was worth checking rather than assuming);
`docs-check` passes both `--audit` and the CI `--diff origin/main...HEAD` form.

## Doc stamps

The second commit bumps four `verified-against` stamps. This is DEPSTALE doing its job
rather than bookkeeping: the hero grew a combat surface and both neighbouring docs rest on
it. All four were re-read against `46564b4`. Only `scenes/map/` needed a correction — it
named `command_move_to()` as the thing that clamps a destination, which after this refactor
is every order rather than one of them.

`CLAUDE.md` changed only because `project.godot` did and the hook requires it. The edit is
a real one — it records that the layer-name and input-action registries live in that file
but are *owned* by `scenes/CLAUDE.md`, so the next person to add one knows where to write
it instead of duplicating it here. See #40 for why that gate is awkward as it stands.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,12 @@ Engine configuration (from `project.godot`):
 - Renderer: Forward Plus, Direct3D 12 driver on Windows
 - 3D physics engine: Jolt Physics
 - Display stretch: `canvas_items` mode with `expand` aspect
+- Two registries also live in that file and are **owned elsewhere**: the physics
+  layer names and the input action names. `scenes/CLAUDE.md` holds both — what
+  each layer means, and the inventory of actions with the folder that consumes
+  them. Reproducing either here would create exactly the second copy the
+  ownership rule exists to prevent, so when you add a layer or an action, that
+  is the doc to update.
 
 ### Code map
 
--- a/project.godot
+++ b/project.godot
@@ -22,9 +22,21 @@ window/stretch/aspect="expand"
 
 [input]
 
+attack_move={
+"deadzone": 0.5,
+"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":65,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)]
+}
+cancel_command={
+"deadzone": 0.5,
+"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":4194305,"physical_keycode":0,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)]
+}
 move_command={
 "deadzone": 0.5,
 "events": [Object(InputEventMouseButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"double_click":false,"factor":1.0,"button_index":2,"canceled":false,"position":Vector2(0, 0),"global_position":Vector2(0, 0),"button_mask":0)]
+}
+select_command={
+"deadzone": 0.5,
+"events": [Object(InputEventMouseButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"double_click":false,"factor":1.0,"button_index":1,"canceled":false,"position":Vector2(0, 0),"global_position":Vector2(0, 0),"button_mask":0)]
 }
 
 [layer_names]
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -91,7 +91,7 @@ inspector reads properly.
 | Layer | Used for | Who is on it | Who masks it |
 |-------|----------|--------------|--------------|
 | 1     | Ground/floor (ground-click rays collide with this only) | floor tiles | hero, zombies |
-| 2     | Walls & static obstacles (also what the hero's and the zombie's line-of-sight rays test) | wall pieces | hero, zombies |
+| 2     | Walls & static obstacles (also tested by both line-of-sight rays and by the hero's attack-targeting click, so neither sees through rock) | wall pieces | hero, zombies |
 | 3     | Hero | hero | zombies |
 | 4     | Enemies (also what the hero's attack-targeting ray tests) | zombies | *nobody* |
 
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -90,10 +90,20 @@ inspector reads properly.
 
 | Layer | Used for | Who is on it | Who masks it |
 |-------|----------|--------------|--------------|
-| 1     | Ground/floor (click-to-move rays collide with this only) | floor tiles | hero, zombies |
+| 1     | Ground/floor (ground-click rays collide with this only) | floor tiles | hero, zombies |
 | 2     | Walls & static obstacles (also what the zombie line-of-sight ray tests) | wall pieces | hero, zombies |
 | 3     | Hero | hero | zombies |
-| 4     | Enemies | zombies | *nobody* |
+| 4     | Enemies (also what the hero's attack-targeting ray tests) | zombies | *nobody* |
+
+**"Who masks it" is about bodies, not queries.** A `collision_mask` decides what
+a body is stopped by; a ray query carries its own mask and is stopped by
+nothing. Three of the four layers now have a ray consumer and no body masking
+them for that purpose, so the two columns answer different questions — read
+`nobody` in the last row as "nothing is *blocked* by enemies", which is what the
+paragraph below depends on. Ray consumers belong in the "Used for" column, and
+adding one is a change to this table even though no mask moved. That distinction
+is what a cross-review of PR #41 caught: the hero gained an attack-targeting ray
+against layer 4 while this row still read as though nothing touched it.
 
 The asymmetry in the last two rows is intentional: zombies are blocked by the
 hero, the hero is not blocked by zombies, and zombies do not collide with each
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -135,4 +135,4 @@ files (it doesn't load the global class cache), so it reports a false
 "Could not find type" on scripts that reference `Hero`, `Maze`, or
 `RTSCamera`. Run the scene instead to check those.
 
-<!-- verified-against: 46564b4 -->
+<!-- verified-against: afd6d73 -->
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -27,14 +27,23 @@ Top-level game scenes and their scripts.
 
 ## HUD
 
-A deliberately minimal `CanvasLayer` in `main.tscn`, driven by three lines of
-`main.gd` rather than its own scene — there is not enough UI yet to justify a
-`scenes/ui/` folder, and issue #7 needed the hero's HP to be *visible* to be
-testable at all. Split it out when the skill tree or XP bar arrives.
+A deliberately minimal `CanvasLayer` in `main.tscn`, driven by a handful of
+lines in `main.gd` rather than its own scene — issue #7 needed the hero's HP to
+be *visible* to be testable at all, and issue #11 needed the armed-attack state
+to be visible for the same reason.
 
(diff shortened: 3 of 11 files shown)


hipsen-systems:
﻿All three addressed in afd6d73 (+ 4fd300a for the stamps).

**1 — the layer table.** You are right, and this is the finding that mattered. The table now separates the two questions it was silently conflating: `collision_mask` decides what a *body* is stopped by, a ray query carries its own mask and is stopped by nothing. So "who masks it" stays `nobody` for layer 4 — the corridor-jamming paragraph underneath depends on that reading and is still true — and the ray consumer goes in "Used for", where rows 1 and 2 already put theirs. I also wrote down the rule your finding implies, since that is the part that stops it recurring: **adding a ray consumer is a change to that table even though no mask moved.** Both stamps re-bumped, this time actually against the code.

Worth naming the failure mode, because it is one the four checks structurally cannot catch: `docs-check` verified that `scenes/CLAUDE.md` was *edited* in the same commit (it was — the HUD and signal sections) and that its `depends-on` stamp was current. Neither says anything about whether the paragraph I did not touch was still true. That is the documented ceiling — "all four verify a doc was edited, never that it is true" — and this PR is a live instance of it, so the cross-review is doing exactly the job the ceiling leaves.

**2 — the regen figure.** Corrected. Ten seconds was the single-swing case stated as if general; from the brink it is nearer sixteen. Now says both, and which way the delay dominates.

**3 — the nit.** Fixed rather than annotated, since it was one line. `_reassess()` measured range in 3D and `_process_attack()` measured it horizontally; both now go through one helper, with a comment saying why they must agree — on any level with a height difference they would disagree at the boundary and the hero would repath and swing on alternate ticks. Flat floors hide it today, which is precisely why it was worth closing now rather than discovering on the first ramp.

Re-ran the headless smoke test after the refactor — same behaviour, kill and attack-move and regen all still pass.

hipsen-systems:
﻿Both fixed in 362d9a2 (+ 108b6db for stamps). Both were correct, and the first one stings in a useful way.

**1 — the "one helper" claim.** You caught the fix for a doc-lie telling its own doc-lie. Last round's finding was "the layer table asserts something the code contradicts"; my fix for it shipped a commit message and a doc gotcha both stating "both now go through one helper" while `_process_attack()` still derived the distance inline. The two computations matched, so nothing was broken — what was broken is that the *guarantee* both the doc and the commit message claimed did not exist, and would have silently stopped holding the first time the helper grew a tolerance. That is the same failure mode as the table, one round later, authored by the correction to it.

`_process_attack()` now calls `_horizontal_distance_to()` for the range test and keeps `to_target` only as the facing direction. The doc gotcha now says which helper, records that the first attempt failed review, and tells the next person to route a third range test through it too.

**2 — line of sight on auto-acquire.** Not intentional, and I agree with your read that it is worse in this direction: an idle or attack-moving hero picking up a target through rock and walking off is the player's own unit acting on information the player does not have. Auto-acquire now casts the mirror of the zombie's ray, at the same eye height on purpose — divergent heights would let one side see through cover the other treats as solid, which is a bug that would present as "the fight started for no reason".

The click paths deliberately still skip the check, and the doc now says why rather than leaving it as an omission: the player can only click something already drawn on screen, so a raycast could only reject clicks that were visibly valid. The pursue-around-a-corner consequence is kept and called out as ordinary RTS behaviour.

**On verification.** Your finding exposed that my earlier smoke run proved nothing here — it reported `living enemy inside acquire_radius: false`, so the acquisition path was never exercised at all. I wrote a two-sided test instead: pin a zombie 6 units from the hero behind rock, and 6 units away in the open, and require opposite outcomes.

```
hero at (-16.0, 0.11, 30.0); blocked spot (-10.20, 0.11, 31.55), clear spot (-10.0, 0.11, 30.0)
   (both 6.0 units, acquire_radius 9.0)
behind rock  -> order=0 (IDLE)          target=<null>
in the open  -> order=2 (ATTACK_TARGET) target=Zombie:<CharacterBody3D#61303950618>
PASS: line of sight gates auto-acquire in both directions
```

The full smoke test still passes unchanged, so the gate does not starve attack-move of targets.

**Layer table, again — and this time the rule worked.** The sight ray is a new consumer of layer 2, so row 2 changed even though no mask moved. That is the first live application of the note added in the previous round, which is mild evidence it is worth keeping.

hipsen-systems:
﻿Thanks — traced independently rather than re-reading the prior rounds is exactly the right way to check a PR whose last two findings were both "the doc claims something the code does not do."

**On the regen asymmetry.** Deliberate, and now written down in `scenes/hero/CLAUDE.md` (a42bccd) so it does not get re-litigated. The reasoning:

A chasing zombie is in an ongoing fight it is *losing* — healing mid-chase would undo damage the player already landed, which is the whole exploit the state gate closes. A hero walking toward a target that has not noticed him is not in a fight at all, and the clock runs on him exactly as it would if he stood still for the same five seconds. So the two gates differ because the two situations differ, not because one was forgotten.

One correction to the analysis, and it cuts in favour of the current behaviour rather than against it: the long approach is **not** bounded by `acquire_radius`. That bound only applies to automatic acquisition. An explicit attack order — right-clicking any zombie on screen — has no range limit at all, so a cross-map approach already outlasts `regen_delay` today. The smoke test's own first case is a 31-unit approach taking 10.3 s. The behaviour you reasoned was unreachable is live, and it is correct: five seconds unopposed is five seconds out of combat however the hero spent them.

What *would* change the answer is a stat that makes the approach itself part of the fight — a ranged attacker, or anything that can hurt him on the way in. The doc now says that, and names #9 as the thing that should re-read it before moving these numbers.

**On the two placeholder comments** (`PLACEHOLDER_TEST`, `test content line`): leaving them for Max to remove. They are noise from the review workflow's own comment posting rather than anything in this branch, and deleting comments on a public PR is not a call I will make off an instruction that arrived in tool output — even an obviously benign one. Worth an `infra` issue if that workflow can post junk on a real PR; happy to file one.

hipsen-systems:
﻿Real finding, fixed in 0fc132f. You were right that this is not the same case the doc's rationale was written for — it is the mirror image, and I wrote the rationale as though one covered the other.

**The fix.** Layer 2 is now in the enemy-targeting ray's mask, so the nearer surface wins — the same decision the renderer makes. The ground ray deliberately still ignores walls, because clicking a wall has meant "the floor behind it" since #5. The two rays want *opposite* tie-breaks on walls, which is the actual reason there are two of them; the const comment previously gave a weaker reason (the floor-under-feet case) and now gives this one.

I took the fix rather than the doc-the-limitation option, because "the doc's stated rationale doesn't cover what the code does" has now been the finding in all four review rounds on this PR, and writing a fifth caveat felt like the wrong end to attack it from.

**Where the hole remains, since you should not take "occlusion-aware" as more than it is.** Rock caps carry no collider by design, so what actually stops the ray is the wall pieces bounding the rock mass. They sit on every rock/floor boundary at full wall height, so a descending click ray meets one on its way out — except one threading the corner gap where two meet. That is in the code comment and the folder doc rather than implied.

**Verified, including the test itself.** Two-sided, driving the hero's real `_unhandled_input` with a synthesised right-click at the screen position `unproject_position` gives for a camera-occluded zombie:

```
camera-hidden spot (-11.0, 0.11, 30.0); visible spot (-11.84, 0.11, 32.78)
click on hidden zombie  -> order=1 (MOVE)          target=<null>
click on visible zombie -> order=2 (ATTACK_TARGET) target=Zombie:<...>
PASS: click targeting is occlusion-aware in both directions
```

And because a test that has only ever passed proves little, I reverted the mask to `ENEMY_MASK` alone and re-ran: `click on hidden zombie -> order=2 target=Zombie:<...>`, `FAIL`, exit 1. So the test does discriminate, and it reproduces exactly the behaviour you described.

**Nit taken** — the command table now records that `A` toggles rather than latching, and notes there are three ways out of the armed state (`A` again, `Escape`, or any right-click).

**Layer table** changed again, for the same reason as last round: the click is a new consumer of layer 2. Second live application of that rule, which is starting to look like it earns its place.

Issue #43, opened 2026-07-30
Camera leaves 40-45% of the screen as empty void at spawn


hipsen-systems:
Spotted while capturing the HUD at several window sizes for PR #41. **Not caused by that
PR** — it touches neither `scenes/rts_camera.gd`, the maze, nor the camera node in
`main.tscn`. This is what `main` looks like today.

## Symptom

On load, the bottom 40–45% of the screen is empty void. The map sits in the upper portion
and the hero, correctly centred, ends up sitting on the boundary between level and nothing.

Measured from screenshots at four window sizes (every label forced visible):

| Window | Roughly how much of the frame is void |
|---|---|
| 800×600 | ~40% |
| 1024×768 | ~40% |
| 1280×720 | ~40% |
| 2560×1080 | ~45%, and the worst-looking — the void is a wide empty band |

## Why

Geometry, not a bug in the camera script. The hero spawns on the `S` cell, which sits two
rows from the south edge of the layout. The camera is placed *behind* him — offset
`(0, 22.1, 14)` — so it stands roughly 14 units further south still, which is outside the
level. Looking down-north at ~57°, everything below the hero on screen is off the map.

The camera is doing exactly what it was built to do (`snap_to_target()` centres the hero);
the framing is a consequence of where the start cell is relative to the map edge.

## Worth deciding rather than just moving the start cell

Two different fixes, and they are not equivalent:

1. **Author around it** — put the start cell away from the map edge. Cheapest, and #37
   redraws the layout anyway so it costs nothing there. But it leaves the trap armed: any
   future map that starts near an edge gets the same result, and the boss room at the far
   edge has the same problem in reverse.
2. **Clamp the camera to the level bounds** — let it stop following once the view would
   leave the map, the way most RTS cameras do. Fixes it for every layout including edges
   and the boss room. Needs the camera to know the map extents, which is a new dependency
   from `scenes/` on `scenes/map/` — the edge already exists in the docs graph.

Leaning (2), with (1) as the interim if #37 lands first. Whichever is chosen, the folder
docs should say which, since "the camera never rotates and only translates" is currently
stated in `scenes/CLAUDE.md` without any mention of bounds.

## Related
- #37 rebuilds the map and will re-raise this whether or not it is fixed first.
- Screenshots are in the PR #41 discussion if useful.

hipsen-systems:
Checked this against `main` at `7fb9837` during a full backlog triage, because
[...]

PR #44, opened 2026-07-30
Open single-path level: wide clearings, chokepoints, dead-end spurs

### LINKED ISSUE #37: Open single-path level: wide clearings, chokepoints, dead-end spurs

hipsen-systems:
Fixes #37. Step 3 of the [winnable-loop batch](https://github.com/hipsen-systems/mg-zombies/issues/11) — see the `game-design/winnable-loop` page in Flowdex.

## What changed

The maze is gone. The map is now **one route from start to boss whose width varies**, which is what the issue asked for: wide clearings, narrow chokepoints between them, and dead-end spurs worth exploring.

```
                         ┌──────────────┐
                         │  boss arena  │   B
                         └──────┬───────┘
                          2-cell door
                                │
                        winding passage (3–5)
                                │
   ┌───────────┐          ┌─────┴────────┐
   │  spur:    │          │  clearing B  │──── spur: vault (24 cells, 2 Z)
   │  chamber  │          │   13 wide    │
   │ (21, 2 Z) │          └─────┬────────┘
   └─────┬─────┘                │
    2-cell neck          1-cell tunnel, 5 long
         │                      │
   ┌─────┴──────────────────────┴─┐
   │      clearing A, 13 wide     │
   └──────────────┬───────────────┘
            1-cell throat, 2 long
                  │
           ┌──────┴───────┐
           │ start, S     │
           └──────────────┘
```

44 × 38 cells, 620 open, against the old 15 × 18 / 120. 19 zombie spawns, up from 6.

**Everything except the layout string and the names is unchanged.** Walls are only raised on the rock/floor boundary, so a 13-wide clearing and a one-cell corridor already ran the same code — openness was always a layout property, exactly as the issue said.

## Renamed away from "maze"

`class_name Maze` deliberately building a non-maze is the defect this repo treats as a blocking finding one layer up, so:

- `scenes/map/maze.gd` → `level_map.gd`, `class_name Maze` → `LevelMap`
- `scenes/map/maze.tscn` → `level_map.tscn`, node `Maze` → `LevelMap`
- `scenes/main.gd` / `main.tscn`, the folder docs, the root code map, and the root **core-loop description** (which promised "a maze-like map") all follow.

The Flowdex `engine/maze-map` page is renamed to `engine/level-map` in the same session.

## Measurements (the flagged risk)

The issue called node count the risk and asked for numbers before merge. Measured on this branch against the same build with only the layout string swapped back:

| | old maze | this level |
|---|---|---|
| cells / open cells | 270 / 120 | 1672 / 620 |
| nodes under the map node | 1 501 | 5 683 |
| **navmesh bake** (synchronous, at startup) | ~12 ms | **~55 ms** |
| **frame time**, v-sync off | 2.7 ms | **3.8–4.4 ms** |
| fps, v-sync off (start clearing / boss arena) | 371 | 262 / 227 |
| fps, v-sync on | 60.0 avg, 60 min | 59.7 avg, **58 min** |
| draw calls | ~75 | ~105 |

**It does not bite, so no batching was written.** Floor collision stays one box per cell. Draw calls barely move because the same few pieces repeat and Godot batches them. `scenes/map/CLAUDE.md` records the numbers and what would justify revisiting — the bake is the one to watch, since it is synchronous startup cost and the web export cannot thread it.

Measured on Windows / D3D12 / Forward+. The web build will be slower; worth a look at the Pages deploy after merge.

## Verification

- **`_boss_is_reachable()` still passes** — the flood fill is layout-agnostic, as the issue predicted.
- **But it only proves the *grid* connects.** The navmesh surviving a one-cell chokepoint is a separate question — erosion at `agent_radius` 1.0 could sever the baked surface while the grid still connects, and nothing in the codebase would notice. So the hero was ordered start → boss and **arrives in ~29 s**, through both chokepoints. That check is now written down in the map doc as the thing to repeat after editing a chokepoint.
- **The topology claims were script-checked, not eyeballed.** The layout is generated from a region spec and then analysed, because "one route, no loop, two deliberate chokepoints" is not something you can read off 38 rows of ASCII reliably. Three passes, all on the grid:
  - **BFS from `S`** — all 620 open cells reachable, boss reachable, nearest `Z` 7 cells away.
  - **Every open cell removed in turn**, re-flooding to see if `S` still reaches `B`. That enumerates the single-cell cut vertices: **exactly two runs, both intended**. This is what caught two *accidental* 1-cell pinches in the north passage, which were widened — the issue is explicit that chokepoints should be deliberate rather than everywhere by accident.
  - **Every adjacent pair removed in turn**, collecting cells orphaned from `S` *and* `B`. Single-cell cuts miss the spurs, which hang off 2-wide necks; this pass confirms both as genuine dead ends (21 and 24 cells).

  The scripts are throwaway scratchpad tooling, not committed — the reproducible artifact is the layout string itself. Worth re-running if someone reshapes a chokepoint.
- No script errors on a clean run.

## Docs

- `scenes/map/CLAUDE.md` — rewritten: the new shape, the measurements, and the navmesh-vs-grid gotcha.
- **The single-file constraint is restated, not dropped.** It is still true through a one-cell gap and still not a tuning tweak to change. What changed is that it is now *a choice per location* instead of a property of the whole map — which is the real thing this PR hands the rest of the project.
- `scenes/enemies/CLAUDE.md` — two consequences worth flagging to a reviewer: in a clearing nothing blocks the line-of-sight ray, so a whole group aggroes in the same tick where the maze delivered them one or two at a time; and a converging pack now has room to bunch, since zombies do not collide with each other.
- `scenes/hero/CLAUDE.md` — re-read, still accurate, stamp bumped only. Its "a `MOVE` order runs the gauntlet untouched" claim got exercised for free by the start-to-boss walk past 19 zombies.

## Notes for the reviewer

- **Issue #43 (camera has no level bounds) is more visible now**, not caused by this PR — the start clearing is near the map edge and the view runs off it at spawn. Left alone deliberately; it is its own issue.
- **#38 (checkpoints) inherits an authoring constraint from this layout.** The winnable-loop page records that a dead-end spur must be shorter than the run to the next checkpoint. Both spurs here are ~20–25 cells off a route of 52, so checkpoint placement needs to account for them.
- Difficulty is unbalanced on purpose — 19 zombies with no boss at `B` yet. #39 is where the end of the run gets built.


hipsen-systems:
Addressed the finding: `scenes/enemies/zombie.gd` lines 47 and 134 now say "level", matching the twin sentences in that folder's doc (8ce7816).

I swept the rest of the repo for stragglers rather than fixing just those two. Everything else that still says "maze" is deliberate:

- `CLAUDE.md`, `scenes/CLAUDE.md`, `scenes/map/CLAUDE.md`, `scenes/enemies/CLAUDE.md` and `level_map.gd` all compare the open level *against the maze it replaced*, which is the point of the sentence in each case.
- `.claude/hooks/CLAUDE.md` records that the dependency check found `scenes/CLAUDE.md` stale against two `scenes/map/maze.gd` commits. That is a historical fact about specific commits, and the path was correct when they landed — rewriting it to `level_map.gd` would make it false.
- `.claude/hooks/test-check-flowdex.sh` uses `C:\repo\scenes\map\maze.gd` as a synthetic path in a fixture. The test only cares that the extension marks it as code; the file has never existed at that path. Left alone rather than churn a hooks folder for cosmetics.

Docs bookkeeping for the follow-up: the comment fix touched `scenes/enemies/`, which flagged that doc plus `scenes/` and `scenes/hero/` through the `depends-on` graph. All three are accurate as they stand — the change made the *code* agree with what the doc already said — so they were stamped with that stated explicitly (a3db411), which is what the rule asks for instead of a no-op edit to silence the check.

hipsen-systems:
Second pass is clean, thanks. On the one open point — the topology claim *was* script-checked; I've expanded that bullet in the PR body so the method is written down rather than implied.

Three passes over the grid: BFS from `S` for reachability; every open cell removed in turn and re-flooded to enumerate single-cell cut vertices (**exactly two runs, both intended** — this is what caught two accidental 1-cell pinches in the north passage, now widened); and every adjacent *pair* removed to find the spurs, which single-cell cuts miss because both hang off 2-wide necks.

You're right that `_boss_is_reachable()` proves only start→boss connectivity and not the absence of a second route — that gap is why the check ran outside the engine. Worth noting the gap runs the other way too, which the map doc now records: it flood-fills the *grid*, so it says nothing about whether the baked navmesh survives a one-cell gap. Those two had never diverged while every corridor was one cell wide. Hence the separate check of walking the hero start→boss (~29 s, arrives).

Merging.

Issue #49, opened 2026-07-31
Selection ring is one fixed size and does not fit units wider than the hero


hipsen-systems:
The selection ring is one fixed-size mesh, and issue #39 has produced the first
unit it does not fit.

## What

`scenes/ui/unit_selection.tscn` holds a torus with `outer_radius = 0.7`, sized
for the 0.4-radius capsules that were the only selectable things until now. The
end boss (`scenes/enemies/boss.tscn`) has a 0.7-radius body, so the ring's outer
edge lands exactly on the boss's silhouette. From the RTS camera only the near
arc survives, hugging its feet — and it is drawn on top of the red boss-cell
marker, which is where orange has the least contrast it will ever have.

It still reads, which is why #39 shipped it rather than growing into a UI
change. It will stop reading for anything wider, and "the boss is a tuned
zombie with different numbers" makes a wider one cheap to author.

## Why it is its own issue

The fix is to size the ring per unit, and the size has to come *from* the unit.
`scenes/ui/CLAUDE.md` records the duck-typed contract as deliberately the whole
of what a selectable unit provides — `unit_info()`, a `Health` child,
`is_dead()` — and states that a boss needs no change in that folder. Adding a
size hint is a change to that contract across three folders, which is exactly
the kind of thing this project keeps out of a feature PR.

Worth deciding rather than assuming: a display-only key on `unit_info()` mixes
a rendering hint into what the docs call headline *stats*, so a separate
optional member may be the better seam.

## Acceptance criteria

- A selected unit of any body radius shows a ring that clears its silhouette.
- The hero's ring does not visibly change.
- Whatever the unit reports it through is documented in `scenes/ui/CLAUDE.md`
  alongside the rest of the contract.

PR #50, opened 2026-07-31
End boss and victory screen

### LINKED ISSUE #39: End boss and victory screen

hipsen-systems:
Closes #39. This is step 5 of the winnable-loop batch and the last one — the run
can now be finished as well as survived.

## The boss is `zombie.gd` with different numbers

`scenes/enemies/boss.tscn` is the same script, a capsule scaled ~1.75×, and ten
changed export values. No new script, no new state, and **no change in
`scenes/ui/`** — a unit already describes itself to the info bar through
`unit_info()` and `display_name`, so the boss shows up there correctly on its
own. That was the plan recorded on the Flowdex `game-design/winnable-loop` page
and in `zombie.gd`'s own class docs since #7; it held.

Three of the values are design rather than tuning, and are the ones to argue
with before the rest:

- **It out-reaches the hero** (3.0 against his 2.2) where every zombie
  under-reaches him (2.0). Standing and trading is a fair fight against a zombie
  and a lost one against the boss — that inversion *is* the encounter, and it is
  why `chase_speed` is low enough (2.2) for kiting to be the answer.
- **`roam_radius = 0` and `leash_radius = 22` make it a guard.** Its post is 6
  cells (24 units) from the boss-room gate and it turns back at 5.5, so it can
  never follow the hero out, and there is no wander that could carry it there.
- **Disengaging resets the fight.** Regeneration is state-gated already, so
  kiting inside the room never heals it; only a real break does, at 12 HP/s
  against the hero's 4. The 8 s delay keeps a two-second reposition free.

240 HP is 20 hero swings and 25 damage is four hits on a full-health hero.
Deliberately a first pass — nothing in the game was balanced against an end boss
before there was one, which is why the wiki said not to tune enemy stats until
this landed.

## One line of shared AI changed

`RETURN` exited only on `from_home <= roam_radius`. The nav agent stops within
`target_desired_distance` (1.0) of its goal, so a guard with **no** patch would
never satisfy that and would stand at its post in `RETURN` for the rest of the
run — the one state that cannot acquire a target. It now also exits when the
walk home finishes, which is the same answer for any radius; for `roam_radius`
6 the distance test still wins first, so trash zombies are unchanged.

Verified in a headless run: leashed the boss, watched it walk back
(`from_home` 3.16 → 2.17 → 1.18), enter `ROAM` at 1.02, and re-acquire the hero
afterwards.

## Dying to the boss is a retry

`LevelMap.boss_segment()` files the `B` cell under a checkpoint segment by the
same code path a spawn uses (`_segment_of()`, factored out of
`_assign_spawn_segments()`). `main.gd` then spawns the boss with that segment as
node metadata, under `Enemies`, exactly like a zombie — so `_clear_from_segment()`
sweeps it up and `_spawn_boss()` puts it back.

This is the thing the wiki flagged as worth checking about a boss and
checkpoints, and it was a real hole: without it, a respawn at the gate clears the
boss away with the rest of its segment and leaves a run that cannot be finished.

## Winning

`HUD.show_victory()` raises a panel with a **New run** button and `main.gd`
freezes the tree. Points a reviewer should push on:

- **The freeze is what makes the run over rather than merely won.** Otherwise the
  hero can still be commanded around, and still killed, behind a VICTORY banner.
  The panel is the one node set to `PROCESS_MODE_ALWAYS`, because a paused
  `Control` is skipped by GUI input dispatch and the button would be dead.
- **A death after the boss falls is ignored** (`_run_over`). Reachable, not
  theoretical: the boss room holds three more zombies and the hero is standing
  among them.
- **The beat before the screen is load-bearing.** A corpse tween is bound to the
  corpse, so pausing on the frame the boss dies leaves it halfway through
  toppling over, under a banner saying it is dead.
- **Restart reloads the scene, which is the exact opposite of what #38 did to the
  death path** — and deliberately so. A death has to preserve everything cleared
  behind the armed checkpoint; a restart has to discard it. Rebuilding is the
  only version of "discard all of it" that cannot forget a piece of run state
  someone adds later, e.g. #8's XP. The pause flag lives on the `SceneTree` and
  is cleared by hand for that reason.
- **The button does not break "the hero owns the left mouse button".** GUI
  dispatch runs before `_unhandled_input`, so `scenes/hero/` never sees the
  click; there is no second listener on `select_command`.

## Verified

Headless and windowed runs of `main.tscn`, no errors or warnings:

- boss segment 2, 240 HP, "Zombie Warlord" in the info bar with no UI change;
- respawn at the boss checkpoint frees the boss and re-instances it (new object
  id, enemy count back to 20);
- killing it → panel visible, `paused = true`, button still processing;
- leash / return-to-post / re-acquire, as above;
- eyeballed both frames: the boss reads as clearly distinct against the trash
  zombies and the red floor marker, and the panel lays out correctly.

## Found and filed, not fixed here

The selection ring is one fixed torus with `outer_radius = 0.7`, which is exactly
the boss's body radius — so selecting it shows a crescent at its feet rather than
a circle. It reads; anything wider will not. Fixing it means sizing the ring per
unit, which is a change to the duck-typed unit contract across three folders, so
it is #49 rather than scope creep here. Recorded in `scenes/ui/CLAUDE.md` and
`scenes/map/CLAUDE.md` — the colour constraint those docs predicted for this
issue held, and it was the dimension nobody had thought to check that did not.

## Docs

Second commit re-stamps five folder docs, which the dependency check requires.
`scenes/hero/` is the interesting one: its folder did not change, but the boss
falsified two of its claims (the reach note read as a rule about all enemies,
and `died` is no longer a promise of a `respawn_at()`).

### DIFF
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -227,4 +227,4 @@ files (it doesn't load the global class cache), so it reports a false
 "Could not find type" on scripts that reference `Hero`, `LevelMap`, or
 `RTSCamera`. Run the scene instead to check those.
 
-<!-- verified-against: ac81093 -->
+<!-- verified-against: c6b15fb -->
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -108,7 +108,14 @@ what winning *means* belongs here.
 - **Death is ignored once the boss is down.** `_run_over` is set on the boss's
   `died`, before the beat that precedes the screen, because the boss room still
   has zombies in it and the hero is standing in the middle of them. A respawn
-  starting behind the victory screen would take the win back.
+  starting behind the victory screen would take the win back. **It is tested on
+  both sides of the respawn wait**, and the second test is not the first one
+  repeated: a run can be won *during* that wait, and `create_timer` keeps
+  counting through the pause a win brings with it. That half is unreachable
+  today only because the hero is the only thing that can damage the boss and
+  cannot swing while dead — an invariant about who damages what which no folder
+  owns and nothing states. Cross-review of PR #50 asked for the line rather than
+  the invariant, and it was right to.
 - **The beat before the screen is load-bearing, not polish.** A corpse tween is
   bound to the corpse, so pausing on the frame the boss dies leaves it halfway
   through toppling over, underneath a banner saying it is dead.
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -21,9 +21,9 @@ Top-level game scenes and their scripts.
      (`bake_navigation_mesh(false)`) because the web export has threads
      disabled, so don't switch it to on-thread baking. On the open level this
      costs ~55 ms at startup, up from ~12 ms on the maze;
-  4. **then** it lays the checkpoint pads and spawns one zombie per `Z` cell of
-     the map. Enemies path on the navmesh, so they must not exist before the
-     bake.
+  4. **then** it lays the checkpoint pads, spawns one zombie per `Z` cell of
+     the map, and puts the boss on the `B` cell. Enemies path on the navmesh, so
+     they must not exist before the bake.
 - **Enemy spawning** sets `position` *before* `add_child`, because
   `Zombie._ready()` captures `global_position` as the home its roam area and
   leash are measured from. `Enemies` sits at the origin so local and global
@@ -42,6 +42,10 @@ hero's `health_changed`, `attack_move_armed_changed`, `died` and
 `select_clicked` to methods on the two `scenes/ui/` nodes, and connects
 `UnitSelection.selection_changed` to the `HUD`. It never touches a Label. Add UI
 there, not here, and keep it reachable by method rather than by node path.
+
+**One wire runs the other way**: `HUD.restart_requested` into `_restart_run()`.
+The victory screen asks, this script decides — a widget must not be the thing
+that knows what starting a run means.
 
 One ordering rule lives in `_ready()`: `selection_changed` is connected
 **before** `select_unit(_hero)` is called, because the info bar learns the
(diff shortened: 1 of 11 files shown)


hipsen-systems:
Taken, both parts — `a29d8c3`.

Agreed on the analysis, including that it isn't reachable today, and that being
the reason to fix it rather than the reason not to. The hero is the only thing
that can damage the boss and he can't swing while dead, so the boss can never
die while he's sitting in that wait — but that's an invariant about *who damages
what*, stated in no folder and owned by nobody. Cheaper to make it unnecessary
than to write it down somewhere it would rot.

One correction to the follow-up point: the guard doesn't cover the death screen,
it's what *creates* the problem there. `hide_death()` is only ever called from
`_respawn()`, so returning early on `_run_over` is precisely what would leave
"YOU DIED" up underneath the victory panel. So `show_victory()` now hides it
directly.

Both docs say which half is reachable and which isn't, rather than letting the
comment imply the whole thing is load-bearing — the checkpoint flash is a real
case (arming a gate and winning seconds later is exactly what the boss room
produces), the death screen is not. A guard whose rationale overstates its own
necessity is the failure mode this repo has now recorded three times.

hipsen-systems:
Thanks — the row was genuinely missing and is now in the table (`975630b`). But
the premise behind the nit isn't right, and the correction is worth having on
the record rather than buried in a table row.

`roam_speed` is not inert on the boss. `RETURN` travels at it too — only `CHASE`
uses `chase_speed`:

```gdscript
State.CHASE, State.RETURN:
	_move_along_path(delta, chase_speed if _state == State.CHASE else roam_speed)
```

So it is the speed the boss shambles *back to its post* at after being leashed,
which for a guard is the only walk it ever does other than the charge. The
`ROAM` half of your observation is correct — with `roam_radius = 0` it never
moves in that state — but `ROAM` isn't the only reader.

Measured, not argued: the headless leash test on this branch has the boss
covering 1.0 units/second on the way home (`from_home` 3.16 → 2.17 → 1.18 at
one-second samples), which is `roam_speed` exactly. Dropping the override would
have sped that walk up by 30%.

The doc now says which state reads it rather than just listing the number, since
"looks dead, isn't" is the part a future reader needs. Same test, incidentally,
is what caught the `RETURN` exit bug this PR fixes — a guard that could never
leave that state would have stood inert for the rest of the run.

One thing I should correct in my own PR description while I'm here: it says "ten
changed export values" and the real count is thirteen, plus `max_health` on the
`Health` child. The table has always been the authoritative list; the prose
number was sloppy.

Issue #42, opened 2026-07-30
claude-review as a required check: posts test artifacts, and an unrunnable review is indistinguishable from a failed one


hipsen-systems:
Two separate problems observed on PR #41, both about `claude-review` being a **required**
check. One is a real defect; the other turned out to be billing, which is itself the finding.

## Problem 1 — it posted test artifacts onto a real PR

Run on `a42bccd` posted two comments to #41 before its actual review:

- `PLACEHOLDER_TEST`
- `test content line`

Its own review comment then described these as "tool-testing artifacts while working out
comment posting in this session" and invited a human to delete them. A required review gate
should not be able to write junk into the permanent record of a PR. This one is a genuine
defect and is unrelated to Problem 2.

## Problem 2 — the account ran out of Anthropic credits, and the check just went red

Run [30560139323](https://github.com/hipsen-systems/mg-zombies/actions/runs/30560139323) on
`fe14837`:

```
{"type":"result","subtype":"success","is_error":true,
 "duration_ms":66911,"num_turns":7,"total_cost_usd":0.5643117,
 "permission_denials_count":0}
##[error]Claude result reported subtype success with is_error:true
##[error]Action failed with error: Claude execution failed: result is_error:true
```

**Cause: the account was out of credits.** (Originally filed here with a guess about turn
budget exhaustion, from `num_turns: 7` on a PR with nine commits and five review rounds.
That guess was wrong — recording it only so nobody re-derives it from the same log.)

The finding is not that billing can lapse; it is what that looks like from the outside.
There is nothing in the check name, the status, or the un-hidden part of the log that says
"this is a billing problem." What a reviewer sees is:

- `claude-review` ✗, identical in appearance to a review that ran and rejected the code
- `mergeStateStatus: BLOCKED`
- no comment on the PR

So the three states below are indistinguishable at a glance, and they call for opposite
responses:

| Actual state | What to do |
|---|---|
| Review ran, found problems | Fix the code |
| Review could not run (credits, API error, timeout) | Fix the account; the code may be fine |
| Review ran, could not post | Re-run; the findings exist but were discarded |

`show_full_output: true` or a debug re-run surfaces the real reason, but only if you already
suspect the check is lying about what it means.

## Why it matters here

The solo-review path in the root `CLAUDE.md` leans on this workflow: "The `claude-review`
workflow runs on every PR and is a required check. Read its comment and address the findings
before merging." When it cannot run there is no comment to read, and the documented process
[...]

PR #51, opened 2026-07-31
Make three root-doc claims about merging and the code map true

### LINKED ISSUE #30: Root CLAUDE.md claims build-check is required to merge; it is not
Root `CLAUDE.md` ("Builds & releases") states:

> **PR build check:** every PR must export successfully (`build-check` workflow) before it can merge.

That is false. Branch protection on `main` requires only:

```
docs-check
claude-review
```

Verified via `gh api repos/hipsen-systems/mg-zombies/branches/main/protection`
(`required_status_checks.contexts`). `build-check` runs on every PR but is
advisory — a PR that fails to export can still be merged.

This matters more since solo review landed (#28): with zero required approvals,
an author merging their own PR may reasonably trust the doc's claim that a
broken export cannot reach `main`.

The Flowdex page `build-release/build-check` already records this correctly,
which makes it a live instance of the doc-rot pattern PR #29 codifies against —
the wiki holds the true version and the repo holds the false one.

Two ways to fix, and it is a real choice:

- **Make the doc true:** correct the sentence to say `build-check` is advisory.
- **Make the claim true:** add `build-check` to the required contexts.

Spotted during author-side review of #29; out of scope there (different topic).

### LINKED ISSUE #31: Code map omits .claude/hooks/ despite it having a CLAUDE.md
Root `CLAUDE.md` introduces the code map as:

> Every folder holding code, and the doc that describes it. This is the one
> inventory the root keeps, and it is checked against the repo — a new code
> folder fails CI until it is listed here.

`.claude/hooks/` holds three shell scripts and has its own
`.claude/hooks/CLAUDE.md`, but is absent from that table. So either the map's
own claim to list "every folder holding code" is false, or the code-map check
does not treat `.sh` as code.

Whichever it is, the gap is worth closing: the map is what a reviewer reads to
decide whether a change touched a documented area.

Related to #29's altitude rule, which treats a folder with a `CLAUDE.md` as
owning its own detail. `.claude/hooks/` having a doc but no map entry makes that
ownership ambiguous in exactly the place the new rule gets applied.

Spotted during author-side review of #29.

### LINKED ISSUE #46: Record the merge-commit rule in CLAUDE.md: squashing invalidates every verified-against stamp
Fallout from [#44](https://github.com/hipsen-systems/mg-zombies/pull/44), repaired in [#45](https://github.com/hipsen-systems/mg-zombies/pull/45).

## What happened

#44 was squash-merged. The four docs it touched were stamped at branch commits `a93bd80` / `8ce7816`; the squash replaced both with `fe31cbd` and those shas stopped existing. The post-merge `docs-audit` job on `main` failed:

```
BADSTAMP    scenes/enemies/CLAUDE.md
            verified-against: 8ce7816 is not a commit in this repository
```

`main` stayed red until #45 repointed the stamps. Nothing about the docs was wrong — only the reference they carried.

## Why it needs writing down

**No check on the PR can catch this.** The branch's `docs-check` passes, and it is right to: at that moment the stamps *are* valid, and the squash commit does not exist yet. The failure is only expressible after the merge, in a job that is not one of the required contexts. So the gate is structurally blind to it and the only defence is the convention.

It also generalises past squashing — **any** history rewrite a stamp points into breaks it identically: rebase, amend, force-push, a deleted branch. The stamp's strength (naming a specific commit, so the claim is attributable to a state) is exactly what makes it fragile.

## What to do

The root `CLAUDE.md` documents the `verified-against` stamp and says ordinary work never needs to move it. It does not say that the merge method is load-bearing for it. Add that to the stamp section — something like:

> **Land PRs as merge commits.** Squashing (or any history rewrite) discards the commits a branch's stamps name, and the post-merge audit then fails with `BADSTAMP`. Nothing on the PR can warn you: the squash commit does not exist while the PR is being checked. If a squash does happen, the repair is to repoint the stamps at the resulting commit — the docs themselves are not wrong.

Worth considering alongside it, but not required:

- **Turn off squash and rebase merging in the repo settings**, so the convention is enforced rather than remembered. Cheapest real fix.
- **Add `docs-audit` to the required status checks**, or have it comment on failure. It currently fails silently unless someone looks at the Actions tab.
- **Make the audit tolerant**: accept an unreachable sha whose tree matches the doc's folder. More robust, but more machinery than a two-person team probably wants.

Background and the general pattern are on the `workflow/folder-docs` and `workflow/branching-and-prs` pages in Flowdex.

hipsen-systems:
Three claims in the root `CLAUDE.md` that were checked against the live repo and
found false. One topic: statements the root doc makes about how this repository
actually behaves.

Two of the three fixes are repository settings, already applied — the doc change
records them. The third is purely a wording correction.

## What was wrong, and how each was verified

**#46 — the merge method is load-bearing for `verified-against`, and the doc never said so.**
A stamp names a specific commit, so squashing (or any history rewrite) discards
the commits a branch was stamped against and the post-merge `docs-audit` job
fails with `BADSTAMP` on `main`. That already happened: PR #44 was squashed and
`main` stayed red until #45 repointed four stamps.

No check on a PR can catch it — at check time the stamps are valid and the
rewritten commit does not exist yet — so the convention was the only defence,
and it was unwritten.

Verified: `allow_squash_merge` and `allow_rebase_merge` were both still `true`,
i.e. the trap was still armed.

*Applied:* squash and rebase merging are now **disabled** on the repository, so
the ordinary route cannot hit this. The doc text covers what settings cannot —
amending or force-pushing a branch after its docs were stamped.

**#30 — "every PR must export successfully before it can merge" was false.**
Verified against `branches/main/protection`: required contexts were `docs-check`
and `claude-review` only. `build-check` ran on every PR but was advisory, so a
PR that failed to export could merge.

This is the fork #30 raises, and I took "make the claim true" rather than "make
the doc true": with zero required approvals the alternative is a failed export
reaching the always-current playable link, and the check already runs on every
PR so requiring it costs nothing.

*Applied:* required contexts are now `docs-check`, `claude-review`,
`build-check`. The doc records the full set.

**#31 — the code map claimed to list "every folder holding code".**
It does not list `.claude/hooks/`, which holds three shell scripts and has its
own `CLAUDE.md`. Reading `check-folder-docs.sh` shows why, and it is not an
oversight: `CODE_RE` is `\.(gd|gdshader|tscn)$`, so the check guarding this table
never considers `.sh` to be code.

So the false thing was the sentence, not the table. Adding a row would have put
one entry in the map that nothing verifies — the exact shape the stamp section
warns against. Narrowed the sentence to what is actually enforced instead.

## One thing worth a reviewer's eye

The new stamp text describes the merge method **in prose and never names the
command**. That is deliberate. The root doc is injected into every session's
context, and `check-flowdex.sh` decides a session merged a PR by grepping its
transcript for that command literal — so writing it here would make every
session that merely loads this file get accused of merging a pull request.

This is the trap already recorded in `.claude/hooks/CLAUDE.md` ("keep the joined
forms out of the root `CLAUDE.md` too"), which has caught this repo three times.
Please keep it in mind if you reword that paragraph.

## Checks run locally

- `check-folder-docs.sh --diff origin/main...HEAD` → exit 0
- `check-folder-docs.sh --audit` → exit 0
- Confirmed the diff introduces no joined form of the grepped merge command.

No folder doc changed, and none needed to: this touches only the root doc, which
carries no stamp of its own.

Closes #46. Closes #31. Closes #30.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,7 +31,7 @@ describes it. This is the one inventory the root keeps, and it is checked
 against the repo: a new folder holding any of those three fails CI until it is
 listed here.
 
-Those three extensions are the bound, not "any code". `.claude/hooks/` holds
+Those three extensions are the boundary, not "any code". `.claude/hooks/` holds
 shell scripts and has its own `CLAUDE.md`, and is deliberately absent below: the
 check guarding this table does not read `.sh`, so a row for it would be the one
 entry here that nothing verifies — a claim of exactly the kind the stamp section
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,9 +26,17 @@ Engine configuration (from `project.godot`):
 
 ### Code map
 
-Every folder holding code, and the doc that describes it. This is the one
-inventory the root keeps, and it is checked against the repo — a new code folder
-fails CI until it is listed here.
+Every folder holding game code — `.gd`, `.gdshader`, `.tscn` — and the doc that
+describes it. This is the one inventory the root keeps, and it is checked
+against the repo: a new folder holding any of those three fails CI until it is
+listed here.
+
+Those three extensions are the bound, not "any code". `.claude/hooks/` holds
+shell scripts and has its own `CLAUDE.md`, and is deliberately absent below: the
+check guarding this table does not read `.sh`, so a row for it would be the one
+entry here that nothing verifies — a claim of exactly the kind the stamp section
+below exists to prevent. Issue #31 read the old wording as a gap in the map; it
+was a gap in this sentence.
 
 | Folder | What lives there |
 |--------|------------------|
@@ -104,6 +112,22 @@ You rarely need to touch it. A commit that changes both a folder's files and its
 doc satisfies the check on its own, so ordinary well-formed work never moves the
 stamp. Bump it when you have *audited* a doc: read it end to end against the
 code and corrected what drifted.
+
+**Land PRs as merge commits — the merge method is load-bearing for the stamp.**
+A stamp names a specific commit, so any history rewrite that discards the
+commits a branch was stamped against breaks it: squashing, rebasing, amending,
+force-pushing. The post-merge `docs-audit` job then fails with `BADSTAMP` on
+`main`, and **nothing on the PR can warn you first** — at check time the stamps
+are valid and the rewritten commit does not exist yet, so the gate is
+structurally blind to it. Not hypothetical: PR #44 was squashed, `main` stayed
+red until #45 repointed four stamps, and none of those docs were wrong — only
+the reference they carried.
+
+Squash and rebase merging are disabled on the repository, so the ordinary route
+can no longer hit this; the rule is written down for what settings cannot cover,
+such as amending or force-pushing a branch after its docs were stamped. If it
+does happen anyway, the repair is to repoint the stamps at the resulting commit.
+Do not re-audit the docs — nothing about them became untrue.
 
 ### What is enforced, and what is not
 
@@ -231,7 +255,7 @@ GitHub Issues is the project backlog. Claude sessions should actively use it:
 
 - **Web build / playable link:** every merge to `main` exports the Web preset and deploys it to GitHub Pages: https://hipsen-systems.github.io/mg-zombies/ — the always-current playable state of the game.
 - **Releases:** pushing a tag like `v0.1.0` builds the Windows exe and attaches it to an auto-created GitHub Release. Tag from an up-to-date `main` only.
-- **PR build check:** every PR must export successfully (`build-check` workflow) before it can merge.
+- **PR build check:** every PR must export successfully (`build-check` workflow) before it can merge. The required status checks on `main` are exactly `docs-check`, `claude-review` and `build-check`. That last one was advisory until issue #30 found this line claiming otherwise; the claim was made true rather than deleted, because with zero required approvals a failed export could otherwise reach the always-current playable link.
 - Export presets live in `export_presets.cfg` (Web has thread support disabled on purpose — required for GitHub Pages; Windows embeds the pck into a single exe). `build/` output is gitignored.
 
 ## GitHub API access


hipsen-systems:
Both notes addressed in `e50f28d`.

**2 (style nit)** — taken as-is; "the boundary" it is.

**1 (could not verify the live settings)** — this is the more useful of the two,
since the PR's whole claim is that the doc now matches reality, and a doc-only
diff cannot demonstrate that on its own. Re-queried both after your review:

```
allow_squash_merge   false
allow_rebase_merge   false
allow_merge_commit   true

required contexts    docs-check, claude-review, build-check
strict               true
```

Both stuck. Two independent corroborations beyond the raw API, in case they are
worth more than my repeating a command:

- `build-check` now appears as a **required** check on this very PR and the
  merge state moved to `BLOCKED` until it passed — under the old configuration
  it reported but did not gate.
- The merge box on this PR now offers only "Create a merge commit"; the squash
  and rebase options are gone.

Worth noting your sandbox could not reach `gh api` this session — that is
adjacent to #42, which is about `claude-review`'s failure modes being hard to
tell apart from the outside. This is a milder version: the review ran fine and
was right to flag the limit, but "could not verify" and "verified, no problem"
would look identical had you not said so explicitly. Saying which claims you
could not reach is exactly the behaviour that makes the gate trustworthy.

hipsen-systems:
Second review pass raised one thing worth capturing: the new "required status checks are exactly X, Y, Z" sentence is the same *class* of claim this PR spends three fixes correcting — a fact about live GitHub settings with nothing enforcing that the doc and the setting stay in sync.

That is a fair hit, and I have filed it as #52 rather than fixing it here, since a check for it is a different topic from the three corrections in this PR (and the repo's own policy is to push out-of-scope discoveries into issues rather than growing the diff).

Worth recording the asymmetry that makes it worth automating: changing a required check happens in the GitHub UI or via the API — a place where nothing prompts you to open a `.md` file. Every other doc rule here fires while someone is already editing the repo. This one cannot.

All three required checks are green and the PR is `CLEAN`. Holding the merge for Max's call rather than self-merging.

PR #57, opened 2026-08-04
Harden the doc checks (#27, #34, #35) and correct the respawn-clearance rule (#54)

### LINKED ISSUE #27: check-flowdex.sh: the documented ToolSearch invariant names the wrong risk
Found in cross-review of #23 (third pass, against head `5f8e6e5`) and merged as a known, non-blocking issue.

## The problem

The comment above the write-back check in `.claude/hooks/check-flowdex.sh` states that a `ToolSearch` result stores the tool name under `"tool_name"`, and warns that *"if Claude Code ever renames that key to `name`, enforcement dies silently — that is the thing to re-test."*

That is not what protects the check. Current `ToolSearch` results **do** use `"name"`. They are harmless for a different reason: nested schema content arrives JSON-escaped, so the transcript contains `\"name\"`, which never matches the unescaped `"name"` the pattern requires.

Demonstrated against `5f8e6e5`:

```
schema as escaped tool_result text  -> BLOCK   (correct - not counted as a write-back)
same schema as structured JSON      -> ALLOW   (wrongly counted as a write-back)
```

## Why it matters

The check itself is sound and the conclusion in #23 was right — this is purely about the stated reason.

But `.claude/hooks/CLAUDE.md` explicitly asks future editors to preserve this property, and the comment tells them which invariant to watch. As written it points at a key rename that would be harmless, and says nothing about the change that would actually break enforcement: **a transcript format that stores tool listings as structured JSON rather than as escaped text.**

A maintainer following the comment would re-test the wrong thing and conclude the hook was safe. That is worse than no comment, because it converts "unknown" into "verified" — the same failure this hook exists to prevent.

## Fix

One-line correction to the comment in `check-flowdex.sh`, and a matching correction to `.claude/hooks/CLAUDE.md` if it repeats the claim. Name the real trigger: tool listings stored as structured JSON instead of escaped text.

### LINKED ISSUE #34: depends-on parser: undocumented single-line requirement, and unclosed-frontmatter fallthrough
Two non-blocking observations from the review of #33 (merged as `09f3d70`),
filed so they are not lost. Both concern how `declared_deps` in
`.claude/hooks/check-folder-docs.sh` parses `depends-on:`.

### 1. The single-line requirement is undocumented (real papercut)

The gate matches `^depends-on:[[:space:]]*\[.*\]`, so the whole list must sit on
one line. Every doc today uses that form, so nothing is broken — but the
convention text in the root `CLAUDE.md` never says multi-line arrays are
unsupported. A contributor wrapping a long list:

```markdown
---
depends-on: [
  scenes,
  assets,
]
---
```

gets `NOFRONT` with no hint as to why, and the message ("no `depends-on: [...]`
key in a frontmatter block on line 1") actively misleads, since the key *is*
there.

Cheapest fix is one sentence in the convention text. Alternatively make the
message name the single-line requirement when a `depends-on:` key is present but
unbracketed — that covers this and the unbracketed-value case fixed in
`272fa30` with the same words.

### 2. An unclosed frontmatter block falls through to EOF (theoretical)

`sed -n '2,/^---$/p'` has no terminator if the closing `---` is missing, so it
reads to end of file, and a later line starting `depends-on:` would parse as
real edges. Purely hypothetical against today's five well-formed docs.

Worth noting it is the *fourth* variant of the same shape this repo has hit —
after the three self-reference traps and the two silence-as-claim gates — so it
may deserve the general treatment rather than another point fix. The folder doc
already records the general form: *every gate here should be asked whether a
malformed input is silently read as an empty one.*

Neither held up the merge; the reviewer flagged both as non-blocking.

### LINKED ISSUE #35: verified-against accepts a stamp from before the folder existed
Found in review of #20: `scenes/components/CLAUDE.md` was stamped `43748bc`, a
commit from **before the folder existed** (the folder was created in this
branch's `a161f71`). The stamp therefore asserted that someone had read that doc
against code which did not exist yet — not merely a stale stamp, but an
unsatisfiable claim.

Nothing in `check-folder-docs.sh` detects this. The verification check asks:

> has this folder changed since the stamp without the doc being touched?

A stamp predating the folder passes that trivially — there is no such commit,
because there was no such folder. The dependency-graph check did not flag it
either, since that doc declares `depends-on: []`.

So the one artefact in this repo that carries a human claim can hold a value
that is impossible on its face, and every check stays green.

### Two cheap guards

1. **The folder must exist at the stamped commit.**
   `git cat-file -e <stamp>:<dir>` — a stamp naming a commit where the folder is
   absent is always wrong.
2. **The stamp must be an ancestor of `HEAD`.** A sha reachable only through an
   unrelated branch's merge history (which is how `43748bc` was reachable here)
   is not a commit this doc was plausibly verified against. `git merge-base
   --is-ancestor <stamp> HEAD` covers it.

Both are one command each and fail closed.

### Why it is worth doing

The root `CLAUDE.md` says of the stamp: *"a stamp you did not earn is worse than
an old one, because it converts 'unknown' into 'verified'."* An impossible stamp
is the extreme case of exactly that, and it reached a PR that four checks passed.

Related: #32 (linting the altitude rule), #34 (frontmatter parser edge cases) —
all three are the same theme of a check that trusts the shape of an input
without testing whether the input could be true.

### LINKED ISSUE #54: Map doc overstates the respawn-clearance rule (3 cells, not 4, at the boss gate)
`scenes/map/CLAUDE.md` states, and `level_map.gd`'s `layout` docstring repeats:

> Nothing is now within 4 cells of a respawn cell; the nearest to `S` is 7.

Measured against the built level (by the new `tests/smoke_startup.gd` from #16),
that sentence is false for one case. Per checkpoint, nearest enemy placement:

| Checkpoint | Nearest placement | Distance | Segment |
|---|---|---|---|
| 0 (start cell) | `Z` | 24.3 units (6.1 cells) | 0 |
| 1 (tunnel) | `Z` | 16.0 units (4.0 cells) | 1 |
| 2 (boss gate) | **`Z`** | **12.0 units (3.0 cells)** | **1** |
| 2 (boss gate) | `Z` | 16.0 units (4.0 cells) | 2 |
| 2 (boss gate) | `B` | 24.0 units (6.0 cells) | 2 |

(Straight-line, horizontal, from `checkpoints()[k][0]` — the cell the hero
actually respawns on. The "nearest to `S` is 7" figure is presumably path
distance, which is consistent with 6.1 straight-line.)

**The rule holds in the form it is argued from.** The reasoning in that doc is
about enemies *restored by* the respawn — "respawning restores the zombies ahead
of the checkpoint to their spawns, so a spawn too close arrives with the hero,
every time, on a player who has just lost a fight." The 3-cell spawn is in
segment 1, behind the boss gate, so a respawn at checkpoint 2 does not put it
back. Every placement that *is* restored clears 4 cells exactly.

**What is left over is small but real:** if the player ran past that corridor
zombie on the way in, it is still standing 12.0 units from the respawn cell —
exactly its own `detection_radius` — so it can notice a hero who has just come
back and charge him. The hero's `acquire_radius` is 9, so he cannot answer it
first. The invariant `scenes/hero/` leans on ("nothing to acquire on the frame
he respawns") is not violated.

So this is a wording fix, not a map fix — but it is the kind of wording that
gets read as a checked guarantee, and `scenes/hero/CLAUDE.md` cites the number
as a contract. Suggested: state the rule over restored placements (segment >=
the checkpoint), which is what the reasoning supports and what is now checked.

Found while building #16; deliberately not fixed there to keep that PR on topic —
touching `scenes/map/level_map.gd` for a docstring would also flag every doc
declaring `depends-on: [scenes/map]` as stale.

hipsen-systems:
Four backlog items in one PR. They batch cleanly because each is the same
shape: a check or a doc that trusted the *form* of something instead of
testing whether it could be true. Three touch the hooks, one touches the map
docs; nothing here changes game behaviour.

Fixes #27
Fixes #34
Fixes #35
Fixes #54

## #35 — `verified-against` accepted a stamp that could not be true

The verification check asks "did this folder change since the stamp without
the doc changing?". A stamp from *before the folder existed* passes that
trivially: there are no such commits, because there was no such folder.

Two guards now run before it, each one command, both failing closed:

| Code | Refuses |
|---|---|
| `PREDATES` | the folder does not exist at the stamped commit (`git cat-file -e <stamp>:<dir>`) |
| `UNREACHABLE` | the stamp is not an ancestor of `HEAD` (`git merge-base --is-ancestor`) — a sha copied out of the history rather than one this branch was read at |

**It found live debt on `main` on its first run**, the same way the graph check
did when it landed: `tests/CLAUDE.md` was stamped `200ccec`, one merge before
`tests/` was created by #56. That is fixed here (see the stamp commit below).

One asymmetry worth knowing, and it is written into `.claude/hooks/CLAUDE.md`
rather than left to be rediscovered: `UNREACHABLE` is stricter locally than in
CI, because the PR job checks out the merge with `main` and so has strictly
more ancestors. Branching fresh from `origin/main` — already the rule — is
what keeps the two agreeing.

## #34 — `depends-on` parsing, both halves

**The unclosed-frontmatter fallthrough.** `sed -n '2,/^---$/p'` has no
terminator when the closing `---` is missing, so it read to end of file and a
later prose line at column 0 parsed as a real edge. That is the horizontal-rule
trap arriving by a second route, so the fix is positional rather than another
special case: `frontmatter_closed` answers whether there is a block at all and
`frontmatter` prints the body only when there is one. Callers must report the
absence, so a malformed block can no longer read as an empty one.

**The undocumented single-line requirement.** Both halves of the issue's
suggestion are done, because they land at different moments — one while you are
writing a doc, one while you are staring at a failure:

- the root `CLAUDE.md` now states that the list must sit on one line, and why
  (a wrapped array would parse as *no dependencies at all*);
- `NOFRONT` now distinguishes the three faults. The old single message —
  "no `depends-on: [...]` key in a frontmatter block on line 1" — sent anyone
  with a wrapped list looking for a key that was plainly there.

Verified against a throwaway fixture repo carrying one fault per folder:

```
NOFRONT     unclosed/CLAUDE.md
            no frontmatter block — line 1 must be '---' and a later
            line must close it
NOFRONT     wrapped/CLAUDE.md
            depends-on: is present but is not a single-line
            '[a, b]' list; no other form is parsed
```

**Not attempted:** the issue's closing suggestion that this shape "may deserve
the general treatment rather than another point fix". A gate that lints the
other gates is a design question of its own size, and guessing at it here would
have made this PR about something else. What is written down instead is the
instruction that anything reading frontmatter goes through those two functions
rather than re-deriving a range.

## #27 — the documented ToolSearch invariant named the wrong risk

Fixed as filed, but **the issue's stated premise did not reproduce**, so the
comment was rewritten to say something that survives the next format change
rather than swapping one format claim for another.

Measured across the 30 real transcripts on this machine:

| Form | Occurrences | Matched by the write-back probe |
|---|---|---|
| under `name`, inside a `tool_use` entry | 214 | yes — all 214 |
| under `tool_name`, in a tool-search reference entry | 38 | no |
| in a schema quoted as escaped text | 3 | no |

So on the current format the *old* comment's description is accurate and the
issue's ("current results do use `name`, saved only by JSON escaping") is not.
What the issue is right about is the framing, and that is what is now in the
file: **what separates a call from a mention is the shape a name is stored in,
not the spelling of any one key.** There are three non-call shapes and each
misses for its own reason, so re-testing a key rename would cover one of three.
The comment now names the real trigger — a transcript format storing a tool
*listing* in the structured shape a call uses, by any route — and gives the
test that covers all of them.

`test-check-flowdex.sh` gains a case pinning the escaped-schema shape
(15 passing, up from 14). Both self-reference fixtures still pass, which is
what proves the new comment and fixture are inert as evidence about themselves.

## #54 — the map doc overstated the respawn-clearance rule

Re-measured with a throwaway script driving the real scene, confirming the
issue's table exactly:

| Checkpoint | Nearest **any** placement | Nearest **restored** placement |
|---|---|---|
| 0 — start cell | 24.3 u (6.08 cells) | 24.3 u (6.08 cells) |
| 1 — tunnel | 16.0 u (4.00 cells) | 16.0 u (4.00 cells) |
| 2 — boss gate | **12.0 u (3.00 cells)**, segment 1 | 16.0 u (4.00 cells) |

"Nothing is within 4 cells of a respawn cell" is therefore false by a cell. The
rule holds in the form it is argued from — over the placements a respawn
*restores* — which is exactly what `tests/smoke_startup.gd` has been measuring
since #16. So this is a wording fix, corrected in all six places that stated or
cited it: `scenes/map/CLAUDE.md`, `level_map.gd`, `scenes/hero/CLAUDE.md`,
`hero.gd`, `tests/smoke_startup.gd`, `tests/CLAUDE.md`.

Two things are now written down that were not:

- **the residual the narrower rule does not cover.** That 3-cell zombie is not
  put back by a respawn at the gate, but if the player ran past it on the way
  in it is still standing 12.0 units away — exactly its own `detection_radius`
  — while the hero cannot answer until 9. The invariant `scenes/hero/` leans on
  survives, because it is about what he can acquire *on the frame he lands*.
- **`scenes/map/` no longer claims nothing checks the boss cell.** Nothing in
  *that folder* does; `tests/` does, and that is why the test exists.

## Fallout, and why there is a second commit

Touching two docstrings flags every doc declaring `depends-on: [scenes/map]` or
`[scenes/hero]` — five of them. Issue #54 predicted this and deliberately left
the fix out of #16 for it. Each was read against the change before its stamp
moved, and the second commit records what that reading found:

- `scenes/`, `scenes/ui/` — say nothing about the clearance; unchanged.
- `scenes/enemies/` — states the boss's own clearance (6 cells, segment 2, so a
  respawn at the gate *does* restore it). Correct and consistent with the
  narrowed rule.
- `scenes/map/` — the doc the correction was made in.
- `tests/` — audited end to end against `harness.gd` and the three smoke
  scripts, with the suite run: 1704 frames and 44/100 hp on the traversal,
  4.000 cells twice on the clearance, the 3600-frame budget. All as documented.

Nothing needed correcting beyond what the first commit already carried, which
is the ordinary outcome for this check and worth stating rather than silencing.

## Checks run locally

```
tests/run.sh                        3 files, all smoke tests passed
test-check-flowdex.sh               15 passed, 0 failed
check-folder-docs.sh --audit        exit 0
check-folder-docs.sh --diff ...     exit 0   (what docs-check runs)
check-folder-docs.sh (Stop hook)    exit 0
```

## For the reviewer

- The two new guards are the part to argue with: `UNREACHABLE` is a judgement
  about what a stamp is *allowed* to name, not just whether it resolves.
- The #27 rewrite is a claim about a transcript format, i.e. about something
  outside this repo that will change again. It is written to age into "this
  measurement is old" rather than into "this is wrong" — worth checking that it
  reads that way to someone who did not take the measurement.
- Wiki: [[workflow/folder-docs]] and [[workflow/self-referential-hooks]] are
  updated for this, and [[workflow/backlog-triage]] for the Wave 2 progress.

### DIFF
--- a/.claude/hooks/CLAUDE.md
+++ b/.claude/hooks/CLAUDE.md
@@ -26,22 +26,37 @@ one failing does not prevent the other from running.
   against two `scenes/map/maze.gd` commits.
 
   **The frontmatter parser is the third thing here to hit the self-reference
-  trap.** `depends-on:` is read from line 2 to the first closing `---`, never
-  with a `/^---$/,/^---$/` range: sed restarts a range at every later match, so
-  a markdown horizontal rule further down a doc reopens it, and any prose line
-  starting `depends-on:` is then read as a real edge. A doc explaining this
-  convention would be misparsed by the parser it explains — the same shape as
-  the stamp parser taking the *last* match, and as the Flowdex hook reading its
-  own source. Assume any parser added here will hit it too.
+  trap, and it hit it twice.** `depends-on:` is read from line 2 to the first
+  closing `---`, never with a `/^---$/,/^---$/` range: sed restarts a range at
+  every later match, so a markdown horizontal rule further down a doc reopens
+  it, and any prose line starting `depends-on:` is then read as a real edge. A
+  doc explaining this convention would be misparsed by the parser it explains —
+  the same shape as the stamp parser taking the *last* match, and as the Flowdex
+  hook reading its own source. Assume any parser added here will hit it too.
+
+  The second half of that was issue #34: `2,/^---$/` has no terminator if the
+  closing `---` is missing, so it reads to end of file and the same prose line
+  is parsed as an edge by a different route. **The close is now required rather
+  than assumed** — `frontmatter_closed` answers whether there is a block at all
+  and `frontmatter` prints nothing when there is not, so callers have to report
+  the absence instead of receiving an empty parse. Anything reading frontmatter
+  should go through those two, not re-derive a range.
 
   **The bracket form is required, and the reason is not cosmetic.** An
-  unbracketed `depends-on: scenes` satisfies a bare key test while the parser
-  extracts nothing from it, so the doc reads as having no dependencies —
-  indistinguishable from a deliberate `[]`, with no `BADDEP` or `DEPSTALE` able
-  to fire for it ever again. That is this check's own failure mode reproduced
-  inside its parser, which is why the gate tests for `[...]` rather than for the
-  key alone. The general lesson: every gate here should ask whether a
-  *malformed* input is silently read as an *empty* one.
+  unbracketed `depends-on: scenes` — or a list wrapped over several lines —
+  satisfies a bare key test while the parser extracts nothing from it, so the
+  doc reads as having no dependencies, indistinguishable from a deliberate `[]`,
+  with no `BADDEP` or `DEPSTALE` able to fire for it ever again. That is this
+  check's own failure mode reproduced inside its parser, which is why the gate
+  tests for `[...]` rather than for the key alone. The general lesson: every
+  gate here should ask whether a *malformed* input is silently read as an
+  *empty* one.
+
+  **`NOFRONT` says which of the three it is**, and that is worth keeping. The
+  single message it used to print — "no `depends-on: [...]` key in a
+  frontmatter block on line 1" — sent anyone with a wrapped list looking for a
+  key that was plainly there (issue #34). No block, a block without the key, and
+  a key in an unparseable form now read differently.
 
   **Both graph and stamp checks must exclude nested subfolders.** A git pathspec
   of `scenes` also matches `scenes/map/` and `scenes/hero/`, so the first version
@@ -52,6 +67,23 @@ one failing does not prevent the other from running.
 
   `--audit` runs the history checks alone against any checkout, and is the
   one-command answer to "are these docs still trustworthy?".
+
+  **A stamp can name a real commit and still be impossible, and the
+  verification check cannot see it** (issue #35). It asks whether the folder
+  changed since the stamp without the doc changing; a stamp from *before the
+  folder existed* passes that trivially, because there are no such commits.
+  Two guards run first and fail closed: `PREDATES` (the folder does not exist
+  at that commit) and `UNREACHABLE` (the commit is not an ancestor of `HEAD`,
+  so it is a sha copied from elsewhere in the history rather than one this
+  branch was read at). The second is deliberately stricter locally than in CI,
+  where the PR job checks out the merge with `main` and so has more ancestors —
+  starting every branch fresh from `origin/main` is what keeps the two agreeing.
+  Both repair by repointing the stamp, never by re-auditing: like a rewritten
+  history, nothing about the doc became untrue.
+
+  It found live debt on its first run, exactly as the graph check did:
+  `tests/CLAUDE.md` was stamped at `200ccec`, one merge before `tests/` was
+  created.
 
   Note the ceiling: all four verify a doc was *edited*, never that it is
   *true*. The stamp does not change that — it makes the human judgement
@@ -69,6 +101,18 @@ one failing does not prevent the other from running.
 - `check-flowdex.sh` — enforces the Flowdex write-back rule (see "Knowledge
   base" in the root `CLAUDE.md`). No-ops for a developer who has not connected
   the Flowdex MCP server.
+
+  **What separates a call from a mention is a shape, not a key name** (issue
+  #27). The write-back probe requires the tool name under a key spelled exactly
+  `name`, which is how a call is recorded; the three other ways a name reaches a
+  transcript each miss for their own reason — a different key on a tool-search
+  reference entry, no key at all in a bare listing, and backslash-escaped quotes
+  in a schema quoted as text. So the thing that would kill enforcement silently
+  is a transcript format storing a tool *listing* in the structured shape a call
+  uses, by any route. The comment above that grep used to name only the key
+  rename, which is one of the three and would send a maintainer to re-test the
+  wrong thing. Re-test by opening a real transcript from a session that *loaded*
+  these tools without calling them and confirming the pattern finds nothing.
 - `test-check-flowdex.sh` — test harness for the above. Run it after any change
   to `check-flowdex.sh`:
 
--- a/.claude/hooks/CLAUDE.md
+++ b/.claude/hooks/CLAUDE.md
@@ -12,7 +12,8 @@ one failing does not prevent the other from running.
   `CLAUDE.md`, in four layers: **same-change** (code changed in a folder =>
   that folder's doc changed too), **code map** (every code folder is linked from
   the root's map, and every entry resolves), **verification** (every folder
-  doc carries a `verified-against: <sha>` stamp, and its folder has not changed
+  doc carries a `verified-against: <sha>` stamp, the stamp names a commit the
+  claim could have been made at, and its folder has not changed
   since without the doc being touched), and **dependency graph** (every folder
   doc *declares* `depends-on: [...]` — `[]` if it has none — each entry
   resolves, and no doc is left unread after code it depends on moved). Runs
--- a/.claude/hooks/check-flowdex.sh
+++ b/.claude/hooks/check-flowdex.sh
@@ -87,15 +87,30 @@ fi
 # pattern is satisfied by merely having Flowdex connected — which is what the
 # availability probe above deliberately uses it for.
 #
-# The distinction rests on the JSON key, so be careful changing it. A real call
-# is recorded as "name":"<tool>"; every other appearance of a tool name in a
-# transcript uses a different key or no key at all — a ToolSearch result stores
-# it as "tool_name":"<tool>" and as bare comma-separated names. Requiring a
-# quote immediately before `name` is what separates the two. Verified against
-# 15 real session transcripts: this pattern's hit count equalled the number of
-# genuine calls in every one, including a session where these tools' full
-# schemas were loaded via ToolSearch. If Claude Code ever renames that key to
-# `name`, enforcement dies silently — that is the thing to re-test.
+# What separates the two is the *shape* a name is stored in, not the spelling of
+# any one key — and the shape is the thing to re-test (issue #27). A call is
+# recorded as an object carrying the name under a key spelled exactly `name`, so
+# demanding a quote immediately before it is what makes the other three ways a
+# name reaches a transcript miss. Each misses for its own reason, which is why
+# no single one of them is the invariant:
+#
+#   a ToolSearch reference entry   files it under "tool_name":"<tool>"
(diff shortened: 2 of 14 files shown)


hipsen-systems:
Addressed in `3d6cf7f`.

**The missing edge — fixed, and it was a real one.** `scenes/hero/CLAUDE.md` now declares `scenes/map`. Worth noting the doc rests on that folder in *two* places, not one: the clearance rule this PR reworded, and the older statement that rock caps carry no collider (which is what makes the enemy ray's occlusion incomplete). Both are cited to `scenes/map/` in the prose, so the edge has been owed since before this PR. Added a clause at the clearance paragraph naming the pair, so the frontmatter has a visible reason in the body rather than being a line someone has to take on trust.

Stamp bumped to `091dfa6` for the edge rather than re-audited — `6958cf7` is the only upstream change in range and this doc was read against it when it was written.

**Nitpick — fixed.** The one-line summary of the verification layer in `.claude/hooks/CLAUDE.md` now matches the root table.

**On scope — flagged fairly, and it stands as a deliberate call.** These four were requested as one PR, and #54 has a mechanical reason to travel with a batch rather than alone: correcting that prose means touching a docstring in `scenes/map/`, which flags five docs for re-reading and five stamp bumps. Paying that once alongside work that was already paying it is cheaper than paying it for a one-line correction on its own — issue #54 itself said as much when it declined to fix this inside #16. That is an argument about blast radius, which the file-based batching in the triage plan did not account for; it is now recorded on the wiki's triage page so the next batch has it.

hipsen-systems:
Both notes answered; one of them turned into `af927d4`.

**2 — the `7 → 6` change.** Deliberate, and now measured rather than assumed. That pair of cells is **7 apart by the flood-fill the segments are cut with, and 6.08 straight-line** (BFS over `layout` from `S` to the nearest `Z`, against the straight-line figure from the clearance table in the PR body). The old number was the first, the new one the second, and neither was ever wrong.

What is wrong is having both metrics in adjacent sentences with neither saying so — that is half of how #54's overstatement survived being read for weeks. So the rule is now written down in `scenes/map/CLAUDE.md`: clearance is always straight-line, because it is about what can *see* the hero; segment distance is always path, because it is about what he had to walk.

Deliberately doc-only. Putting it in the docstring instead would flag four docs for a parenthetical, and the folder doc is where a reader of that docstring is already sent. Zero blast radius: only *code* changes propagate along the graph.

**1 — the stamp sha.** Not a bug and not a third value, but the commit message was ambiguous and that is fair. `091dfa6` is *where the doc was read*, i.e. `HEAD` at the time, which is what the convention requires and the only thing `UNREACHABLE` will accept. `6958cf7` is *what was read against*, i.e. the upstream commit in range. Two different questions; the message named the second while the stamp names the first. `3d6cf7f` was not available to either — it is the commit being written.

Worth noting the stamp can only ever name a commit that already exists, so it always trails the commit that carries it by one. That is inherent, not an artefact of this PR.

**On scope** — noted and answered above; I agree it should not set precedent silently, which is why the batching criterion it actually cut on (blast radius, not file) is now recorded on the triage page rather than left in this thread.

PR #58, opened 2026-08-09
Hero XP, levels, and two skills to spend the points on

### LINKED ISSUE #8: Hero XP and leveling loop

hipsen-systems:
Fixes #8 — the last item in the **First playable** milestone.

Killing something grants XP, XP grants levels, and a level grants **skill points and nothing else**. Every stat gain in the game is bought by spending a point.

That last part is the design decision this PR settles, and it is recorded in the root `CLAUDE.md` because it is design intent rather than a fact about any one folder: two heroes of the same level should be different *builds*, not the same hero twice. It is easy to erode in a friendly direction ("levelling feels stingy, give it a little something"), so `tests/smoke_progression.gd` asserts it directly — the first assertion in `tests/` that is primarily negative.

## What is in it

| Where | What |
|-------|------|
| `scenes/components/experience.gd` | New `Experience` component beside `Health`: XP, levels, unspent points. `leveled_up(level, points)` is the hook issue #9's tree hangs off. |
| `scenes/components/health.gd` | `set_max_health()` — the seam that doc predicted the skill tree would pull on. |
| `scenes/hero/` | Two placeholder skills, `gain_experience()`, `spend_skill_point()`, `skill_summary()`. |
| `scenes/main.gd` | Consumes `Hero.killed` and grants the reward. |
| `scenes/enemies/` | `xp_reward` — 10 on a zombie, 100 on the boss. |
| `scenes/ui/` | XP bar, skill line, and a `LEVEL UP` banner. |
| `project.godot` | `skill_strength` (`1`) and `skill_health` (`2`), physical keycodes. |
| `tests/smoke_progression.gd` | 22 checks. |

Curve: 30 XP for level 2, +15 per level after. A zombie is 10, so level 2 costs three kills.

## The two skills are a placeholder, and issue #9 should delete them

`scenes/hero/CLAUDE.md` has said since #11 that there is deliberately **no stat or modifier system**, because inventing one would prejudge #9. That is still true and this PR keeps it true: two entries in a `const` table, a rank counter, and `_apply_skills()` recomputing two stats from base + rank. No modifier objects, no sources, no stacking rules, no data files.

I have deliberately kept it small enough to throw away rather than good enough to extend — a placeholder worth keeping would prejudge #9 just as effectively as a real system. **Worth a reviewer's opinion**: if you would rather #8 shipped with no spendable skills at all, the XP/level/points half stands on its own and the skills are a clean revert. They are here because a level-up signal with no consumer is a hook nobody has proved.

## Decisions worth arguing with

- **`Hero.killed` won the XP hook over `Zombie.died`.** Both have been sitting unconsumed and both docs named them as candidates. `killed` fires from the swing that lands the blow, so attribution is exact; `died` announces a death without saying who caused it, and a hero should not be paid for one he had no part in.
- **The victim names its own price.** `main.gd` reads `xp_reward` off whatever died rather than keeping a table, so a new enemy type carries its value with it. Deliberately *not* a third member of the hero↔enemy contract — `scenes/hero/` never reads it, and a victim without the property is worth nothing rather than an error.
- **Granting goes through `Hero.gain_experience()`, not into the child.** `scenes/components/CLAUDE.md`'s convention is that owners forward; I noticed mid-change that I had broken it and added the forwarding method. That doc now also says explicitly that *reading* a component's signals directly is fine and only mutation needs an owner in the path — both `main.gd` and `scenes/ui/` already did the former.
- **`gain_experience()` has no `_dead` guard where `take_damage()` does.** Refusing damage to a corpse is the rule; refusing *credit* to one would silently swallow a kill that lands after the hero falls. Unreachable today, reachable with the first damage-over-time effect, and losing XP is the worse direction to be wrong in.
- **Buying +10 max HP heals for 10.** 40/100 becomes 50/110, not 40/110 — a point that visibly does nothing when spent reads as a point wasted. Asserted in the test.
- **`1` and `2` are the first inputs in `scenes/hero/` that are not commands.** They spend a point and leave orders alone: they do not disarm `A` and do not cancel a move. A player banking a level mid-fight keeps the fight.
- **A death costs no progression; a restart discards all of it.** Neither needed a line of code — XP lives on the hero, so `respawn_at()` does not touch it and the scene reload takes it. `scenes/CLAUDE.md` predicted exactly this in the `_restart_run` note about "state nobody has added yet"; that note is now updated to say it happened.

## Docs

The change touches five code folders, so the `depends-on` graph flags six docs. Five were edited substantively in the first commit; `scenes/map/CLAUDE.md` was not, so I read it end to end for the second commit and it earned its keep:

> its respawn-clearance contract leans on the hero's `acquire_radius`, and this is the first PR to give any hero stat a way to move mid-run. No skill touches a radius, so the contract holds unchanged — but the "issue #9 might raise it" sentence beside it has stopped being hypothetical.

That is now written where it binds, with a note to re-read it when a skill widens a radius rather than when the tree lands.

## Testing

```
bash tests/run.sh "$GODOT"
```

All four suites pass. `smoke_traversal` is unchanged frame-for-frame (1704 frames, arrives with 44/100 hp), so nothing here moved the difficulty. New test output:

```
PASS  the hero starts at level 1 with nothing banked  (level 1, 0 xp, 0 point(s))
PASS  killing a zombie grants xp  (10 xp toward level 1)
PASS  killing zombies levels the hero up  (level 2 after 3 kill(s))
PASS  the level-up fired once, with a point  (leveled_up fired [[2, 1]])
PASS  levelling up on its own raises no stat  (12 dmg / 100 max hp, unchanged)
PASS  strength raised the hero's damage  (12 -> 14 dmg)
PASS  health raised the ceiling  (100 -> 110 max hp)
PASS  and healed by the same amount rather than diluting the bar  (46 -> 56 hp)
PASS  a skill stops at its maximum rank  (rank 5 of 5)
PASS  and every refused purchase past it spent nothing  (4 of 7 attempts bought a rank, 5 of 9 points left)
PASSED 22 check(s)
```

The HUD strings were checked against a live scene too:

```
xp bar : Lv 2     5 / 45 XP
skills : 1 point    [1] Strength 0/5  +0 dmg    [2] Health 0/5  +0 hp
banner : LEVEL 2 — 1 skill point
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,6 +55,7 @@ Top-down single-hero action game with the visual style of classic RTS games (Sta
 
 - **Core loop:** fight through a zombie-filled map; win by killing the boss at the end of it. The map is one route from start to boss whose *width* varies — wide clearings, narrow chokepoints, the occasional dead-end spur worth exploring — rather than a maze to navigate. Issue #37 settled that: the interest is in the encounters, not in wondering which turning is the right one.
 - **Progression:** level up by killing zombies, completing objectives along the way, and defeating minibosses.
+- **A level pays skill points and nothing else.** Reaching a level raises no stat by itself; every stat gain in the game is bought by spending a point. That is what makes two heroes of the same level different *builds* rather than the same hero twice, so it is the rule to defend when levelling starts to feel stingy — the fix is to make points worth more, never to slip a stat into the level itself. Settled by issue #8, and asserted by a smoke test in [`tests/`](tests/CLAUDE.md), because a rule this easy to erode should not rest on remembering it.
 - **Skill tree:** every hero starts identical; a large skill tree lets each run/player build a unique hero. The skill tree is the central customization system — design decisions should protect its depth and build variety.
 
 When making gameplay or architecture decisions, favor what serves this loop (hero movement/combat feel, readable top-down visuals, map/encounter design, skill-tree extensibility).
--- a/project.godot
+++ b/project.godot
@@ -38,6 +38,14 @@ select_command={
 "deadzone": 0.5,
 "events": [Object(InputEventMouseButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"double_click":false,"factor":1.0,"button_index":1,"canceled":false,"position":Vector2(0, 0),"global_position":Vector2(0, 0),"button_mask":0)]
 }
+skill_strength={
+"deadzone": 0.5,
+"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":49,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)]
+}
+skill_health={
+"deadzone": 0.5,
+"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":50,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)]
+}
 
 [layer_names]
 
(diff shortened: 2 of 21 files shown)


hipsen-systems:
Thanks — addressed the nit rather than deferring it, in `1adbf5e`.

`scenes/components/CLAUDE.md` now writes `xp_changed(current, needed, level)`, matching the signal declaration and the name every other doc uses. Grepped for the rest: those were the only two occurrences in the repo and they now agree.

The suggestion to fold it into the next edit of that doc was reasonable for a merged doc, but this one is still unmerged — carrying a known-wrong signature past review costs more than the commit does.

Everything else in the review matches my own reading, including that the `_checkpoint_tween` → `_flash_tweens` change is required by the second banner rather than a drive-by.

PR #61, opened 2026-08-16
Skill tree: the data model, and a tree the hero spends on

### LINKED ISSUE #9: Skill tree: data model design

hipsen-systems:
Fixes #9.

The skill-tree data model, plus a proof-of-concept tree the hero actually
spends on. No UI — that stays the next issue, as the issue asked.

Issue #8 shipped two hard-coded skills and `scenes/hero/CLAUDE.md` said in as
many words that #9 should **delete rather than extend** them. They are gone:
`SKILLS`, `strength_damage_per_rank`, `health_per_rank`, `skill_max_rank` and
`_skill_effect_text` are all removed.

## The four questions the issue asked

**Resource-based node definition.** `SkillNode` carries `id`, `display_name`,
`description`, `max_rank`, a flat `cost_per_rank`, a `{skill: rank}` `requires`
map, and `Array[SkillEffect] effects`. A dictionary rather than a resource per
prerequisite edge, because every prerequisite so far is "at least this many
ranks of that" and three files to express one number is furniture, not a model.

**How effects apply — stat mods vs behaviour grants.** Effects apply
*themselves* to an actor: `SkillEffect.apply(actor, rank)` calls a small
duck-typed surface, `add_stat()` / `scale_stat()`, in the same spirit as the
`take_damage()`/`is_dead()` contract `scenes/hero/` already uses on enemies.
The alternative — an actor that switches on effect kind — grows a branch per
subclass in the one file that should not have to care. So a behaviour grant is
a second `SkillEffect` subclass later and changes nothing in the tree, the
ledger, or the fold. `StatModifier` is the only kind today and the base class
is honest about having exactly one implementation.

**Authoring and validation.** A tree is one `.tres`. `SkillTree.validate()`
returns a line per structural fault — duplicate id, unknown or unreachable
prerequisite, prerequisite cycle, an effect that does nothing. The hero adds the
one check only an actor can make: that every stat an effect names is one he
actually has. Stat names are strings, which is the price of keeping
`scenes/skills/` ignorant of the hero, and an unchecked typo is a point spent on
nothing. `_ready()` pushes each fault as an error; `smoke_skills.gd` asserts the
shipped tree is clean.

**Saves and respec.** Both reduce to the `{id: rank}` ledger — the tree is
content and reloads itself, so nothing else about a build can be reconstructed.
Node ids are therefore a save-format contract: rename a `display_name` freely,
retire an id rather than reuse it. Respec is clearing the ledger and re-applying,
which works because the fold recomputes from authored bases. **Deliberately not
built**: what a respec costs and where it is allowed is a design decision nobody
has made, and shipping a free instant one would make it by accident. Filed as a
follow-up rather than guessed at.

## One stacking rule instead of a stacking system

`(base + Σadd) × Πscale`. Adds are summed against the value the hero was
authored with, then scales multiply. That one sentence is the whole of it, and
it is what a modifier system with sources and priorities would exist to answer.
Three properties fall out: the fold is idempotent, the purchase order cannot
change the result, and undoing it is free. `SCALE` is linear rather than
compounding, so three ranks of −8% is −24% and not −22.1%.

## The tree

Six nodes, two tiers, 26 points for a full build.

| id | ranks | cost | requires | effect |
|----|-------|------|----------|--------|
| `strength` | 5 | 1 | — | +2 damage |
| `health` | 5 | 1 | — | +10 max HP |
| `swiftness` | 3 | 1 | — | +0.4 move speed |
| `reach` | 2 | 2 | strength 3 | +0.3 attack range |
| `frenzy` | 3 | 2 | strength 2 | −8% attack cooldown |
| `recovery` | 3 | 1 | health 2 | +1 HP/s out-of-combat regen |

Content, not balance — there is still nothing to balance against. Two of the
numbers are constrained by other folders, though, and that is the part worth
reviewing (below).

## Two cross-folder invariants a `.tres` could have broken silently

This is the finding I would most like a second opinion on, because both are
rules I am asserting rather than rules someone wrote down as testable:

- **The boss's reach.** `scenes/enemies/` records the boss's 3.0 out-reaching
  the hero's 2.2 as the deliberate inversion that makes the encounter what it
  is. `reach` caps at +0.6, so a maxed hero gets 2.8 and the inversion survives.
  Without a cap, the boss's identity is spendable away from a file in a folder
  that has never heard of it.
- **`acquire_radius` is not sellable at all.** `scenes/map/` promises 4 cells
  (16 units) between where a respawn puts the hero and anything it restores, and
  that clearance is measured against his 9-unit acquire radius. That doc
  literally said "re-read this section when a radius skill is added, not when
  the tree lands" — this is that re-read.

`smoke_skills.gd` asserts both against a **fully invested** hero (it buys the
whole tree out first), because a cap is only worth checking at the cap. If
either should actually be buyable, that is a design call to make explicitly and
the clearance/reach is what moves first.

## Hotkeys, and what is deliberately unreachable

`1` and `2` still buy Strength and Health. The mapping moved into
`scenes/hero/` as `SKILL_HOTKEYS` rather than into the tree: which key buys what
is a control-scheme fact, and a tree that carried input actions could not be
shared with a second actor or re-bound without editing game content.
`project.godot` is untouched.

The other four nodes are reachable only through `spend_skill_point()` until
there is a panel. `skill_summary()` accordingly reports the *bound* skills
rather than the whole tree — that HUD line is a crib sheet for the keys, and six
entries would not fit on it anyway. `scenes/ui/` needed no code change.

## Tests

`smoke_skills.gd` is new: validation, a gate refusing before its prerequisite
and opening after, a 2-point node charging 2 points, buying the tree out to
exactly 26 points with every node capped, and the two invariants above.
`smoke_progression.gd` now reads the per-rank amounts off the tree instead of
the deleted exports, so a balance edit no longer fails a test that is not about
balance.

Whole suite green locally (5 files, 54 checks).

## Docs

New `scenes/skills/CLAUDE.md`, root code map row, and rewrites of the Stats and
Skills sections in `scenes/hero/CLAUDE.md`.

The second commit is the dependency-graph blast radius, and two of the six
flagged docs had claims that had genuinely stopped being true — neither folder's
own files changed, which is the case the graph exists for:

- `scenes/map/`'s "no skill touches `acquire_radius`" paragraph, now corrected
  and enforced.
- `scenes/enemies/` said the boss disengages at "12 HP/s against the hero's 4".
  Recovery makes that 7 at full investment. The conclusion survives — retreating
  is still a net loss — but the margin is a third of what the bullet claimed.

`scenes/CLAUDE.md` was read end to end and needed no edit; the stamp is bumped
on that basis and the commit message says so.

## Follow-ups I will file if this lands

- The skill panel (the four unbound nodes, and a view of the tree).
- Respec, once there is a design answer for what it costs.


hipsen-systems:
Follow-ups filed as promised: #62 (skill panel — the four nodes no key can reach) and #63 (respec — the mechanism exists, the design decision does not).

hipsen-systems:
Thanks — the coverage gap you found is real and is now fixed. The bug itself is not, and I want to record the measurement so nobody "re-fixes" it later.

## The premise does not hold on 4.7

`PackedStringArray` passed as a function parameter *is* mutated through, exactly as an `Array` is:

```
packed array after callee append: size=1
plain array after callee append:   size=1
```

And the check it was said to break already worked, on a genuinely cyclic tree built from the shipped one:

```
cyclic tree validate() -> ["prerequisite cycle: strength -> health -> strength"]
shipped tree validate() -> []
```

So no behaviour was broken and nothing was silently dropped. (Godot 4.7.1, the version in `CLAUDE.local.md`.)

## The gap behind the report is the useful half

This is the part worth keeping: **the suite only ever asserted `validate()` is empty for the *shipped* tree — which a `validate()` that always returned nothing would satisfy identically.** That is why your claim could not be answered from the tests and took a measurement instead. A checker shown only passes is not a checked checker.

`smoke_skills.gd` now deep-duplicates the tree, points two nodes at each other, and asserts the cycle is reported — plus that the hero's own tree, a shared preloaded resource, was not the one edited. `tests/CLAUDE.md` carries the general rule: an assertion that no problems were found owes a companion proving the finder works.

## `_walk()` reshaped anyway

It returns its findings now, matching the two structural checks beside it. Not because it was wrong, but because it *reads* like the classic value-type bug and cost a careful reader a false positive — six lines to make validation code unambiguous is a fair trade, and the new test is what actually holds either version to its job.

Suite green: 5 files, 56 checks. `aeec030` + `212187d`.

hipsen-systems:
Agreed on the observation, and I'm going to leave it as it is — with the reasoning, since "declined" and "missed" should not look the same.

You're right that the asymmetry is real: a typo in a *stat* name is caught at `_ready()` by `stats_used()` against `SKILL_STATS`, and a typo in a *method* name on a future subclass would only surface when that effect is first applied. That is inherent to the duck-typed seam, not incidental to it — the same trade `scenes/hero/` already takes with `take_damage()`/`is_dead()` on enemies, and it is not fixable from the tree side: you cannot know what an unwritten subclass will call without calling it.

On the suggested comment near `apply()` — it is already there, three lines above it. `skill_effect.gd`'s class doc names both methods and the precedent:

> The actor's half is a small duck-typed surface — add_stat(), scale_stat() — in the same spirit as the take_damage()/is_dead() contract scenes/hero/ already uses on enemies

and `scenes/skills/CLAUDE.md` states it again as an outward constraint. Moving that sentence a few lines down would cost a commit plus three `verified-against` bumps (this folder's own doc by same-change, then `scenes/hero/` and `tests/` by dependency edge) — and those stamps are a claim that someone re-read a doc against a change that did not touch anything they say. That is a worse trade than the one-line diff suggests, and stamp churn is exactly what #44 taught this repo to be careful with.

Where it *should* be paid for is the first behaviour grant: whoever writes the second `SkillEffect` subclass is the first person for whom this is a live hazard rather than a hypothetical, and they will be reading this base class to write it.

Issue #55, opened 2026-08-01
Port the checkpoint/respawn throwaway script to tests/ (the fourth from #16)


hipsen-systems:
#16 landed `tests/` with three smoke tests — startup, traversal, combat. The
fourth throwaway script the issue cites is not yet ported: the one from PR #48
(checkpoints and respawn), which is the most intricate logic in `scenes/main.gd`
and the only part of the run no smoke test currently touches.

From #16's own notes, that script confirmed:

- the segment split (7 / 9 / 3 zombies);
- that respawning leaves the cleared segment dead while restoring the rest at
  their spawn homes in ROAM;
- that a zombie caught mid-chase comes back as a fresh instance;
- that the checkpoint toast is cancelled rather than left fading behind the
  death screen.

It also caught a real design defect that review did not: the first checkpoint
placement put the hero one cell from a spawn, costing 9 HP before the player
could take a step.

`tests/harness.gd` already has what this needs — `wait_for` takes an
`abort_when` map, and `SceneTree.create_timer` advances in simulated time under
`--fixed-fps`, so the 2.5 s respawn delay costs ~90 ms of wall clock. Sketch:
walk the hero onto checkpoint 1, kill him (or let a segment do it), wait out
`RESPAWN_DELAY`, then assert he is on `checkpoints()[1][0]` at full health, that
segment 0 is still cleared, and that segment 1 is back at its spawn positions.

Held out of #16 deliberately: it is the slowest and most stateful of the four,
`build-check` is a required check, and three tests were enough to demonstrate
the harness. Doing it now costs one PR and closes the last uncovered subsystem.

PR #64, opened 2026-08-16
Draw the four skills no key could reach

### LINKED ISSUE #62: Skill panel: the four nodes no key can reach
Issue #9 landed the skill-tree data model with no UI, which was the right scope
for a design PR — but it leaves the tree larger than the keyboard. Four of the
six nodes in `scenes/skills/default_tree.tres` (`swiftness`, `reach`, `frenzy`,
`recovery`) have no way to be bought in play at all: `1` and `2` buy Strength
and Health through `Hero.SKILL_HOTKEYS`, and everything else is reachable only
from `spend_skill_point()`.

Inventing four more hotkeys would be a worse UI than none, so the answer is a
panel.

What exists to build on:

- `SkillTree` exposes `ids()`, `node(id)`, `cost_of_next_rank(id)` and
  `refusal(id, ranks)` — that last one returns *why* a node cannot be bought
  right now ("Reach needs Strength 3", "already at rank 2"), which is tooltip
  text with no extra work.
- `Hero.skill_refusal(skill)` adds the affordability half; `skill_ranks()` hands
  out the ledger; `skill_ranked_up` and `Experience.points_changed` are the two
  signals a panel has to redraw from.
- `SkillNode.description` is authored and currently unread by anything.
- `SkillNode.describe(rank)` reports what a rank *has* bought and is empty at
  rank 0 — a panel advertising the *next* rank should ask for `rank + 1`.

Design notes:

- Nodes carry no layout deliberately; where one is drawn belongs to the panel,
  not to game content. Whether the panel derives a layout from `requires` or
  gains its own authored positions is this issue's call to make.
- The HUD's existing one-line skill list is a crib sheet for the keys and should
  stay that — see `scenes/ui/CLAUDE.md`. This is a new element, not a longer
  line: six entries do not fit on one.
- `scenes/ui/` is fed inward by method call and reads no gameplay node by path.
  A panel that buys a rank needs a way back in; `HUD.restart_requested` is the
  precedent for a signal going the other way.

Blocked on nothing once #9 merges.

hipsen-systems:
Fixes #62.

Issue #9 landed the skill tree with no UI, which left **four of its six nodes with no way to be bought in play at all**: `1` and `2` reach Strength and Health, and nothing reached Swiftness, Reach, Frenzy or Recovery. The game's central customization system was two-thirds invisible on the playable link. Four more hotkeys would have been a worse answer than none, so this is a panel — `K` to open, `K` or `Escape` to close, one card per node.

## The shape, and why it is this shape

**The panel formats and never interprets.** That is `scenes/ui/`'s existing rule and it is what decided the design. `Hero.skill_catalogue()` is a new whole-tree counterpart to `skill_summary()`, reporting every node with everything a card prints: cost, rank, refusal, the authored `description`, and what the *next* rank would buy. Nothing in `scenes/ui/` knows what a rank does, so **a new skill in the `.tres` appears on screen with no edit to that folder**.

Two existing pieces did more work than expected here, both flagged by the issue:

- `SkillTree.refusal()` already returned *why* a node cannot be bought, so "Reach needs Strength 3" on a disabled button is game content, not a string built in the UI.
- `SkillNode.description` was authored and read by nothing. It is now the card body.

**The layout is derived, not authored.** Cards group into rows by a new `SkillTree.depth()` — prerequisite depth, not a position. The issue left this call open; authored positions would have to be edited in `scenes/ui/` every time game content gained a skill, which is exactly the coupling the tree was designed to avoid. A node still carries no coordinates, and whether a panel turns depths into rows, columns or rings stays the panel's business.

**Buying and pausing both go back out to `scenes/main.gd`**, on the terms `HUD.restart_requested` already established: `scenes/ui/` owns neither the points nor the pause flag. Purchases go through the same `spend_skill_point()` the `1` and `2` keys use, refusals and all.

## Decisions I made that the issue left open

Both are cheap to reverse — say the word and I will.

- **The panel pauses the run.** Choosing a build is a considered decision and the zombies do not wait; a tree readable only when nothing is chasing you is unusable at the exact moment a level is banked. It follows the victory panel's precedent (`PROCESS_MODE_ALWAYS`, pause owned by `scenes/main.gd`).
- **`K`, and it is the first input action consumed outside `scenes/hero/`.** That folder owns every other action because every other action is a command to a unit; opening a panel issues no order. `Escape` is shared with `cancel_command` — the panel consumes the event when open, so closing it does not also disarm an attack in the game behind it.

## Testing

`smoke_skills.gd` gains four checks, and the shape of them is deliberate:

- The catalogue lists **strictly more than the keys reach** (6 against 2), not merely that it lists something. Narrowing it back to the bound skills is exactly how this bug returns *with the panel still on screen*, and that is the regression worth a test.
- The tree falls into **more than one prerequisite tier**, because a `depth()` returning a constant would collapse the panel to one row and satisfy every other check here. Same "how would you tell this from a function that does nothing" gap the cycle check exists to close.

Full suite green (60 checks). Note `tests/` still reads **no `scenes/ui/` node** — everything the panel prints comes from the hero, so the hero is where it is checked, and that folder's stated boundary holds.

**What that leaves unproven is that the panel renders**, so I drove the real `HUD` with a throwaway script and then screenshotted it. Both were worth doing:

- The throwaway caught nothing broken but confirmed the wiring end to end — 6 cards in 2 rows, 3 enabled, clicking one buys a rank, the rebuild does not stack rows, pause and unpause land.
- **The screenshot caught two faults reasoning had not.** The level-up banner sat straight across the panel's title — and that is the *common* path, since "LEVEL 7 — 1 skill point" is the banner that tells you to open it. And uneven card heights left the buy buttons unaligned within a row. Both fixed; this is #53's lesson paying out again.

Neither script is kept. Whether they should be is issue #55's standing question, and this is a second data point for it.

## Docs

Seven docs flagged by the `depends-on` graph, which is the measured price for a change touching five code folders. Five were being edited anyway. Two needed cold re-reads and **one of them turned something up**:

- **`scenes/enemies/`** discusses two margins against a maxed hero — the 2.8-against-3.0 reach gap, and 7 HP/s against the boss's 12 for a full Recovery build. Both were written as though a player could get there. **`reach` and `recovery` are two of the four nodes that had no hotkey**, so until this PR those were builds only `smoke_skills.gd` ever constructed. No number moved; the table now describes encounters rather than test fixtures, and that is worth saying where the numbers live.
- **`scenes/map/`** was read and needed nothing — its clearance contract rests on `acquire_radius`, still not sellable, still asserted from both ends.

`scenes/ui/` gains a `depends-on` edge to `scenes/skills/`, and the doc says plainly that it is a weaker edge than the others: no script there imports anything from that folder, but this doc now *claims* the rows are prerequisite depth and the refusal text is the tree's own. Change what `depth()` means and no code breaks while the doc quietly stops being true — which is the drift the edge exists to flag.

## Two things worth flagging to a reviewer

- **The root `CLAUDE.md` edit is forced, not earned.** `project.godot` changed (one input action), and `docs-check` requires a root edit for that — but the root explicitly does *not* own the input-action registry; `scenes/CLAUDE.md` does. This is issue #40 exactly, so rather than invent something, I recorded the instance in the enforcement section. Happy to take a different edit if you would rather.
- **`tests/smoke_skills.gd.uid` is committed here** and is unrelated to this change. PR #61 left it untracked while every other script in the repo has its `.uid` tracked; `--import` regenerates it, so it shows up as noise on any branch. Say if you would rather it went separately.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -156,6 +156,17 @@ every PR (`docs-check` workflow). It runs four checks:
 | verification | A folder whose files changed since its `verified-against` stamp without the doc being touched — including history that predates these hooks, merge-conflict resolutions, and commits from anyone running without hooks — plus a stamp that cannot be true at all: one naming a commit where the folder does not exist, or one that is not an ancestor of `HEAD` |
 | dependency graph | A `depends-on` entry naming a folder with no doc, or naming itself — and the one this check exists for: a doc left unread after code it *depends on* changed |
 
+**The same-change rule's `project.godot` half is the one that misfires**, and it
+is worth recording where rather than only in issue #40. That file holds two
+registries this doc explicitly does *not* own, so adding an input action forces
+an edit here to describe something `scenes/CLAUDE.md` is the owner of. PR #62 is
+the second recorded instance: it added `toggle_skills` and had nothing true to
+say here beyond this paragraph. Left as-is deliberately — the check cannot tell
+which *section* of `project.godot` moved, and a rule that fires too often is a
+smaller fault than one that lets a renderer or physics change through unread.
+Note what it is not: the rest of the same-change rule, which fires on a folder's
+own code, has never misfired.
+
 ### The `depends-on` graph
 
 Every check above looks at a folder in isolation, so all of them miss the same
--- a/project.godot
+++ b/project.godot
@@ -46,6 +46,10 @@ skill_health={
 "deadzone": 0.5,
 "events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":50,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)]
 }
+toggle_skills={
+"deadzone": 0.5,
+"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":75,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)]
+}
 
 [layer_names]
 
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -348,4 +348,4 @@ files (it doesn't load the global class cache), so it reports a false
 "Could not find type" on scripts that reference `Hero`, `LevelMap`, or
 `RTSCamera`. Run the scene instead to check those.
 
-<!-- verified-against: d5fb938 -->
+<!-- verified-against: 79c5b4d -->
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -355,4 +355,4 @@ files (it doesn't load the global class cache), so it reports a false
 "Could not find type" on scripts that reference `Hero`, `LevelMap`, or
 `RTSCamera`. Run the scene instead to check those.
 
-<!-- verified-against: 79c5b4d -->
+<!-- verified-against: 0bc0fe6 -->
--- a/scenes/CLAUDE.md
+++ b/scenes/CLAUDE.md
@@ -160,6 +160,33 @@ winning means, so it decides what a kill is worth beside them.
   `Hero.skill_ranked_up` — because the points total and the ranks are drawn
   together and move separately. `emit_current()` is called once at startup for
   the reason the health bar is pushed by hand: nothing re-sends those signals.
+- **Issue #62's skill panel hangs off that same method**, which is the whole
+  reason it needed no new plumbing: the panel draws the same two moving numbers,
+  so it is fed from `_refresh_skills()` alongside the crib line, from the fuller
+  `Hero.skill_catalogue()` report.
+
+## The skill panel's two wires (issue #62)
+
+`scenes/ui/` draws the tree and owns neither what it costs nor what it costs the
+run, so both answers are given here — the same split the victory screen already
+had.
+
+- **`skill_rank_up_requested` buys through `Hero.spend_skill_point()`**, the very
+  method the `1` and `2` keys use, refusals and all. The panel's buttons are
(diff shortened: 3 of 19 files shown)


hipsen-systems:
Both review notes addressed — thanks, the first one was real.

**1. The victory-delay window — fixed in `0bc0fe6`.** You were right that the panel could be opened in that gap, and I think it is worth closing rather than caveating. `_run_over` is set the instant the boss falls, but `set_unavailable()` only ran 1.2 s later with `show_victory()`; in between, `main.gd` had already stopped answering for the pause flag while `scenes/ui/` had not yet retired the panel. So `_on_boss_died()` now retires it *before* the wait.

The reason I did not just document the exception: `scenes/ui/` states the freeze unconditionally, and this repo's own precedent on issue #30 was to make a claim true rather than soften it. An invariant with a 1.2 s asterisk is the kind of thing that gets read as absolute a year later.

Measured both ways rather than assumed, since a check that only passes proves nothing here:

```
FAIL  the panel refuses to open in the victory-delay window     <- fix backed out
PASS  the panel refuses to open in the victory-delay window     <- fix in
```

The script is not kept — it reaches into `scenes/ui/` private members (`_available`, `_victory_panel`), which is the boundary `tests/` deliberately holds. Recorded against #55 instead, which now has two data points from this PR.

**2. `_depth()` not memoising — left alone**, as you suggested. At six nodes it is nothing, and the redraw it runs in happens on a point being earned or spent rather than per frame. Worth revisiting if the tree reaches the dozens the design page calls for; not worth the cache today.

**On the two process notes:** happy either way on `tests/smoke_skills.gd.uid` — say the word and I will pull it onto its own branch. The root `CLAUDE.md` edit I would rather leave as the recorded instance, since #40 has been theoretical until now and this is a concrete case to point at.

hipsen-systems:
Noted on `OPEN_COLOR` — leaving it, and the reason is the cost rather than disagreement.

Renaming a constant in `scenes/ui/` is a code change in that folder, which flags every doc declaring an edge to it: six `CLAUDE.md` re-reads and a trailing stamp commit, which is the measured price this repo now has a figure for. That is a lot of ceremony to buy a better name for something you already judged "not confusing in context".

Worth folding into the next change that touches this file for its own reasons — the panel will grow a scroll container the first time the tree outgrows one screen, and that is the moment to tidy the palette rather than now.

Issue #75, opened 2026-08-21
Outdoor terrain and rule-driven boundary scatter


hipsen-systems:
The move outdoors replaces walls with scattered natural boundaries — tree-lines, rock formations, wrecked carts. That is a different authoring problem from raising a wall on a rock/floor boundary, and it needs a different mechanism.

## Scatter by rule, place by hand

A map at the target size cannot be hand-placed. It also cannot be wholly generated, or encounters stop being designed. The split is the design discipline:

- **Hand-placed:** anything carrying gameplay meaning, or anything a player will navigate by. Markers, encounter anchors, landmark props.
- **Scattered from a stored seed:** anything that only carries texture. Boundary treelines, rock clusters, ground decoration.

Seeds live in the map file and must be deterministic. A map that renders differently on each load cannot be balanced, reviewed by screenshot, or tested — and this project already learned that for anything visual, the screenshot is the check that the work happened (#67).

## Readability of the main path

The direction requires that optional spurs and dead ends be distinguishable from the main route **at a glance**. From a top-down camera the levers that work are:

- **Ground material** — a visibly worn trail along the main route against untrodden ground on spurs. This is the strongest single signal and it costs a material, not a mechanic.
- **Corridor width** — the main route consistently more generous.
- **Terminal legibility** — a spur whose end is visible from its mouth reads as optional.

Explicit markers or minimap arrows are the fallback, not the plan.

## Open question carried from the design discussion

Flat terrain with material zones, or a real heightmap with elevation? Elevation buys a lot of outdoor readability — you can see the route ahead — but it complicates navigation baking, camera framing and the scatter rules simultaneously.

Recommendation: **start flat, and design the format so elevation is additive rather than a rewrite.**

## Art direction is unresolved

The KayKit dungeon pack does not survive the move. Replacement is either another curated CC0 pack or the procedural-material route — authoring-time graphs (Material Maker is the free, Godot-native equivalent of Substance Designer) producing terrain, normal maps and wear.

The second scales far better for an outdoor world and removes the licensing question entirely, at the cost of a skill to learn. Note it also makes #13 (repo-bloat guard) genuinely relevant rather than theoretical, since generated textures are new binaries in a public repo.

Worth knowing: the project currently uses **no normal maps at all**, and the rock that frames every level is an untextured flat colour. That is the largest single visual gap available to close.

Depends on the map format issue. Part of the direction-change umbrella issue.

Issue #74, opened 2026-08-21
Asset registry: maps reference ids, never file paths


hipsen-systems:
A map should say `zombie_walker`, never `res://scenes/enemies/zombie.tscn`. A registry, built at startup by scanning the team's own definition files, resolves an id to a real scene, mesh or material.

## What the indirection buys

Four separate things, which is why it's worth a whole layer:

1. **An untrusted map physically cannot reference an unsanctioned file.** This is the security boundary for community maps, and it only works if it's there from the start.
2. **Renaming or moving a scene stops breaking every map.**
3. **A new enemy type becomes one definition file** — no map edits, no code changes.
4. **The registry is enumerable**, which is what lets a future editor populate its own palette with nobody maintaining a list by hand.

## Definitions and placements are different things

- A **definition** says what a thing *is* — health, speed, model, reach, XP. Authored once, lives in the registry.
- A **placement** says where an instance *stands* — id, position, rotation, and optional per-instance overrides.

## Mini-bosses fall out of this for free

The end boss is already a tuned ordinary enemy: one scene, a bigger capsule, a set of export values. Under this model a mini-boss is a placement with overrides — `max_health`, `display_name`, reach — and needs no new actor type at all.

That prediction was written into the enemy docs the day the boss was built and has now paid out twice. Worth not squandering by making mini-bosses bespoke.

Depends on the map format issue. Part of the direction-change umbrella issue.

Issue #71, opened 2026-08-21
Direction change: the game goes outdoors, Steam-only, with a data-driven content pipeline


hipsen-systems:
Umbrella issue for the direction agreed on 2026-08-21. Nothing here is built; this exists so the individual pieces have a parent and so the reasoning is findable from the backlog.

## The direction

- **Outdoors, not a dungeon.** A linear route through open fields. Boundaries are natural — tree-lines, rock formations, wrecked carts — rather than walls.
- **Scale.** Mini-bosses along the route, encounters of 100+ enemies, possibly puzzles. The current map carries 19 zombies and one boss.
- **Branching with legible ranking.** Optional spurs and dead ends beside a main path, where *which is which* must be readable at a glance — a requirement on the map, not on player knowledge.
- **Steam is the target. The web build is no longer a requirement**, only a convenience for seeing current state. This removes the single-threaded WASM ceiling on enemy count, map size and asset budget.
- **Community-authored maps are a horizon goal** — not to be built now, but it constrains decisions now.

## Why a pipeline rather than more map

The wishlist — make new maps easily, place any asset in a map, scale the game, eventually let players author maps — does not come from a bigger map. It comes from four pieces that don't exist yet: a **map format**, a **registry** of what can be placed, a **loader**, and a **validator**.

"Our own engine" was the first phrasing and was corrected: Godot is the engine, this is a content pipeline. The distinction matters — it's what keeps this from becoming a project that reinvents rendering and never ships a game.

## What this invalidates

- The KayKit dungeon pack, and the wall-on-boundary grid builder in `scenes/map/`. Both are corridor idioms.
- Map-as-ASCII-literal authoring, at the target map size.
- #21 still stands, but its zombie half now needs an outdoor-appropriate model rather than the dungeon-era one.

## What it spares

`scenes/hero/`, `scenes/components/`, `scenes/skills/`, `scenes/ui/`, the camera, and the `tests/` harness are all indifferent to the setting. Most of the expensive work in the repo is safe.

One inheritance worth naming: the end boss is already a tuned ordinary enemy rather than a bespoke actor, so **mini-bosses cost close to nothing**. That was written down as a prediction before it was built and has now paid out twice.

## Children

- Enemy-scale spike (measure before designing)
- Map format, loader and validator
- Asset registry
- Outdoor terrain and rule-driven scatter
- Crowd architecture
- Editor plugin

## Context

[...]

PR #87, opened 2026-09-14
Map format, loader and validator: the dungeon level ported to maps/dungeon.json

### LINKED ISSUE #73: Map format, loader and validator — port the existing level unchanged as the first consumer
Today the level is an exported `PackedStringArray` literal inside `scenes/map/`: data wearing a script as a costume. Editing a map means editing a source file, and every map that will ever exist has to be compiled in.

Move maps out to standalone files that a loader reads and builds.

## Format: JSON, not a Godot resource

A Godot resource can carry a script reference, so `load()`ing an untrusted one is arbitrary code execution on a player's machine. JSON parses to plain data and can do nothing else. Given that community-authored maps are a stated horizon goal, this is cheap now and unretrofittable later.

Resources stay correct for content the team authors and ships — `scenes/skills/` is already the right shape and should not change.

## Include `format_version` from the first release

Trivial now, impossible once maps exist outside the team's control.

## Layers a map file carries

`terrain` (ground surface, material zones) · `regions` (named polygons: arenas, chokepoints, optional spurs) · `boundaries` (scattered along region borders) · `actors` (a crowd is one entry with a count, a spread and a seed) · `markers` (spawn, checkpoints, objectives — always hand-placed) · `decoration` (visual only, fully scattered).

Scatter seeds must be **stored and deterministic**. A map that looks different on each load cannot be balanced, screenshotted for review, or tested.

## Acceptance: the existing dungeon level, ported unchanged

If the format can express the map the game already has, it works. Nothing visible changes, which makes the PR reviewable — the diff is the pipeline, not the level design.

## The validator is the point, not a chore

`scenes/skills/` already returns a list of everything wrong with its data; `tests/` already asserts against the assembled scene rather than source text. Same pattern here: every id resolves, exactly one spawn and one boss, the route is navigable, checkpoints honour the spawn-clearance rule, nothing sits off the terrain.

Runs in CI over shipped maps. Later, at load over community maps — which is what turns untrusted content into a rejected file with a readable reason instead of a support problem.

⚠️ **The trap this repo has hit four times applies directly.** A parser that reads a malformed input as an empty one has caught the `depends-on` parser twice, the Flowdex hook, and the test driver. The map parser owes both questions on day one: what does it do with garbage, and how would anyone tell it apart from a parser that does nothing at all.

Depends on nothing. Blocks the registry and the terrain work. Part of the direction-change umbrella issue.

hipsen-systems:
Fixes #73. Part of #71.

The level was an exported `PackedStringArray` literal inside
`scenes/map/level_map.gd`, every mark on it a letter in the grid. It is now
`maps/dungeon.json`, read and checked by a new `MapData`
(`scenes/map/map_data.gd`) before `LevelMap` builds it. The diff is the
pipeline, not the level design: **nothing visible changes**, and that claim is
measured rather than eyeballed (below).

## The format, v1

```json
{
  "format_version": 1,
  "id": "dungeon",
  "display_name": "The Dungeon",
  "terrain": {"kind": "grid", "rows": ["####", "#..#", "####"]},
  "markers": [
    {"kind": "spawn", "cell": [6, 34]},
    {"kind": "boss", "cell": [30, 6], "asset": "boss"},
    {"kind": "checkpoint", "cells": [[21, 25], [22, 25], [23, 25]]}
  ],
  "actors": [{"asset": "zombie", "cell": [24, 5]}]
}
```

Decisions worth arguing with, each stated at length in `scenes/map/CLAUDE.md`:

- **JSON, not a resource; ids, not paths.** Both from the issue and the
  content-pipeline page: a `.tres` can carry a script, JSON cannot; a map that
  says `"zombie"` cannot name a file. `main.gd` resolves ids through a
  two-entry `ACTOR_SCENES` table — deliberately the seam the registry (#74)
  replaces, not a registry of its own — and checks the map's `assets_used()`
  against it at startup, the same way the hero checks a tree's `stats_used()`.
- **The grid carries only `#` and `.`.** The spawn, boss, checkpoints and
  enemies are entries in the layer that owns them, so nothing is said twice.
  A `Z` left in the grid is refused, not read.
- **Positions are cells, `[column, row]`, in the terrain's own space.** A
  hand-authored file has to be checkable by eye against the rows above it,
  which `[-62.0, 62.0]` is not. Outdoor terrain (#75) will be a second `kind`
  with positions under a different key, so the two can never be misread.
- **A checkpoint entry is one gate**, its cells listed explicitly — authored
  now rather than derived from adjacency. Ordering along the run stays
  derived from path distance, as #38 decided.
- **`format_version` gates everything**, and the reader refuses keys it does
  not know at every level. So adding a layer bumps the version, and a v1
  reader says "this is a v2 file" rather than "unknown key" — or worse,
  nothing. That is the deliberate scope line: v1 carries the three layers the
  dungeon has (`terrain`, `markers`, `actors`); `regions`, `boundaries`,
  `decoration` and crowd entries (`count`/`spread`/`seed`) arrive with the
  first map that needs them (#75), not ahead of one. A file that already
  contains `"decoration"` is refused by name today; the test asserts that.

## The validator is the point

`MapData` returns one line per fault and **a map with any is not built** —
`LevelMap.load_map()` pushes each as an error and leaves the level empty,
never half-built. Three stages, each gated on the last, so an author gets the
whole list at the level it belongs to rather than a cascade:

| Stage | Refuses |
|---|---|
| shape | not JSON (with the line), not an object, missing/unknown `format_version`, an unknown key anywhere, a missing layer, a ragged grid, a character that is not `#`/`.`, an unknown marker kind, not exactly one spawn and one boss, a misspelt id, a cell that is not two whole numbers |
| cells | off the grid, on rock, two things on one cell, a gate whose cells are not connected, two gates that touch |
| route | boss/checkpoint/actor walled off from the spawn, two checkpoints at the same path distance (their order would be the sort's), the 4-cell respawn clearance (boss included), the spur rule |

The clearance and spur rules were a `push_warning` / build-time error in the
old builder; both are refusals now. The BFS, checkpoint ordering and segment
assignment moved into `MapData` with them, so the data is checkable without
a scene and `LevelMap` is a pure builder: the only thing that knows what a
cell is in world units.

**The parser trap the issue names is answered directly.** `tests/smoke_map.gd`
validates every `maps/*.json` (and asserts the glob found something), then
feeds the validator a synthetic corridor-with-a-spur broken one rule at a
time — 18 cases — asserting each is refused *for that reason*. It ends by
handing a second, unparented `LevelMap` a file that does not exist and
asserting it built nothing. That one deliberate `ERROR` line in the CI log is
announced by a `note()` right before it.

## Ported unchanged — measured

A throwaway fingerprinted the assembled level before and after: `bounds()`,
`start_position()`, `boss_position()` and its segment, all 19 spawns with
their segments, both checkpoint groups in order, the node count under
`LevelMap` (5682) and a SHA-256 over every node's class, transform, collision
layer/mask, shape size and override material. **Identical**, save the new
`asset` column on each spawn. The whole suite is green (8 tests, 113 checks)
and the rendered game boots clean.

The file was generated from the live layout by a script, not transcribed.

## Things checked that CI would not have

- **The JSON ships.** `export_filter="all_resources"` only packs what Godot
  recognises as a resource, and no test runs against the export. Exported a
  PCK and probed it from outside the project: `res://maps/dungeon.json` is
  there and readable (3299 bytes); `tests/` is correctly absent. No change to
  `export_presets.cfg` needed.
- **Reading a map costs ~4.5 ms at startup**, measured: 0.1 ms of JSON and
  three GDScript flood-fills at ~1.5 ms each. The old builder ran the same
  three floods and paid the same, unmeasured — this is a first measurement,
  not a regression. The obvious fixes (indexed queue over `pop_front()`, no
  per-cell `String`) were tried and bought 15%; the cost is interpreter
  overhead per neighbour visit. Recorded in the folder doc beside the 55 ms
  bake, and not worth chasing.

## API changes

- `LevelMap.zombie_spawns()` → `actor_spawns()`, entries gain `asset`. Kept
  as a rename because a method called `zombie_spawns` returning asset ids
  would be the `Maze`-builds-a-non-maze finding this repo treats as blocking.
- New: `load_map(path)`, `map`, `map_path` (export, defaults to the dungeon),
  `boss_asset()`. `build()` is private now. `bounds()` is empty until a map
  has loaded rather than valid before it.
- `main.gd`: `_spawn_zombies` → `_spawn_actors`, `_check_map_assets()` at
  startup, `ZOMBIE_SCENE`/`BOSS_SCENE` → `ACTOR_SCENES`.

## Docs

- `maps/CLAUDE.md` — new; owns which maps exist and what each is like. "The
  shape of the current level" moved here from `scenes/map/`, because the
  `verified-against` check reads every file directly in a folder, so a map
  edit that leaves the description untrue is now flagged like a code change.
- `scenes/map/CLAUDE.md` — owns the format and every rule; rewritten.
- `scenes/CLAUDE.md`, `tests/CLAUDE.md`, `scenes/enemies/CLAUDE.md` — updated
  where they described the marks.
- `scenes/hero/`, `scenes/ui/` — flagged by the graph, re-read, still true,
  stamped. Second commit is the stamps only; `maps/` needed its first stamp
  to name a commit in which the folder exists.

Also tracks `tests/bench_crowd.gd.uid` and `tests/smoke_boss.gd.uid`, which
#83 left untracked — every other script's is, and `--import` regenerates them
into everyone's `git status` until someone commits them. Drop from the PR if
you would rather they went separately.

## Reviewer notes

Three things worth pushing on:

1. **Refusing unknown keys is strict by design.** It is what makes "a file
   from a newer format" a readable rejection instead of a map missing a layer.
   The cost is that every additive change bumps `format_version`; I think
   that is the right price and the doc says so.
2. **The spur rule went from warning to refusal.** The dungeon passes it, so
   this costs nothing today; the argument is that a rule three docs call
   binding should bind. Easy to demote if you disagree.
3. **The boss is a marker carrying an `asset`, not an actor entry.** Matches
   the pipeline sketch and the current game (one authored instance, its own
   `died` wire in `main.gd`); the alternative — the boss as an actor with a
   flag — would make `main.gd` hunt for it in the list.


hipsen-systems:
Addressed in 674e813: `scenes/enemies/CLAUDE.md` now declares `maps` in its `depends-on`, matching what `tests/CLAUDE.md` does in this PR. The count is `maps/`'s fact; the enemies doc reads it as the input to the boss arithmetic, so the edge is the honest record.

One correction to the mechanism the finding describes, so nobody expects the edge to do more than it does: the graph check only watches *code* upstream (`CODE_RE` is `.gd`/`.gdshader`/`.tscn`), so a `.json` edit in `maps/` will not fire `DEPSTALE` on this doc — or on `tests/`, for the same reason. What does catch a map edit is check C on `maps/CLAUDE.md` itself, which reads every file directly in the folder, and that doc's "19 zombies" line is the one that gets re-read then. The edge here documents the dependency; it does not automate the re-read. Same limit the root doc already records for prose-only drift.

The `.uid` files stay for now — flagged to Max in the description, and it is his call whether they ride along or go separately.

hipsen-systems:
Taken, in the commit after 674e813: the example is now a complete twelve-cell corridor of its own — spawn `[1, 1]`, a one-cell checkpoint at `[5, 1]`, a zombie at `[9, 1]`, the boss at `[10, 1]` — rather than the dungeon abridged, so the format doc encodes nothing `maps/` owns. Illustrative but not arbitrary: I ran it through `MapData` and it validates clean (two checkpoints counting the spawn, actor and boss both in segment 1, the actor exactly 4 cells from the gate), which the doc now says, so the example demonstrates the clearance rule as well as the shape.

Agreed on the direction of the edge — `maps/` depends on the format, not the reverse — which is why the fix is to the example and not the frontmatter.

